PR #2, opened 2019-06-18
Gatsby setup

OskarAhl:


gatsby-config.js
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+	siteMetadata: {
+		title: `Deriv.com`,
```
ashkanx: Eventually, we should have different title & description for each page. so I think this can be removed.

OskarAhl: Yes I'll check how we can make this dynamic + include translation for the description

PR #3, opened 2019-06-19
Add websocket handling

OskarAhl:


src/common/websocket/config.js
```diff
@@ -6,64 +6,62 @@
  * git update-index --assume-unchanged src/javascript/config.js
  *
  */
-const domain_app_ids = {
-    // these domains as supported "production domains"
-    'deriv.com': 11780,
+const domain_config = {
```
ashkanx: I think production & staging don't matter here, as users won't be able to log in. also, there will be the same redirect url.

src/common/websocket/config.js
```diff
@@ -6,64 +6,62 @@
...
 
 const getSocketURL = () => {
     let server_url = window.localStorage.getItem('config.server_url')
     if (!server_url) {
-        const loginid = window.localStorage.getItem('active_loginid')
-        const is_real = loginid && !/^VRT/.test(loginid)
-        const server = isProduction() && is_real ? 'green' : 'blue'
-
-        server_url = `${server}.binaryws.com`
+        server_url = `blue.binaryws.com`
```
ashkanx: let's connect to default or green

src/components/layout/header.js
```diff
@@ -35,7 +36,7 @@ const Header = ({ siteTitle }) => (
                     padding: `0.5rem`,
                 }}
             >
-                Home
+                <Localize i18n_text="Home" />
```
ashkanx: what an ugly syntax 😃

src/common/websocket/config.js
OskarAhl: @ashkanx The app id is used for the websocket connection and we need the hostname to match `app_id` with environment

src/common/websocket/config.js
OskarAhl: ok 👍

src/components/layout/header.js
OskarAhl: Got a more beautiful suggestion? 😅 Either way can change it later when we add translations

zakibinary: Maybe just `text` instead of `i18n_text`?

OskarAhl: @zaki yeah `text` is cleaner but I think we want to keep it more specific incase we need a extract translations script?

ashkanx: `translate`?

PR #13, opened 2019-07-02
Amam/modal fix

cakasuma:
- fix breakpoints
- fix modal components

### DIFF
--- a/src/components/containers/portal.js
+++ b/src/components/containers/portal.js
@@ -2,22 +2,16 @@ import { Component } from 'react'
 import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
 
-// Use a ternary operator to make sure that the document object is defined
-const portalRoot =
-    typeof document !== `undefined` ? document.getElementById('modal') : null
-const gatsbyRoot =
-    typeof document !== `undefined`
-        ? document.getElementById('___gatsby')
-        : null
+// Check these element is not null before accessing them
+const has_document = typeof document !== 'undefined'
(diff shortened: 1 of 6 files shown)


src/components/containers/portal.js
```diff
@@ -0,0 +1,53 @@
+import { Component } from 'react'
+import ReactDOM from 'react-dom'
+import PropTypes from 'prop-types'
+
+// Use a ternary operator to make sure that the document object is defined
+const portalRoot =
+    typeof document !== `undefined` ? document.getElementById('modal') : null
```
OskarAhl: can we add a `const has_document = typeof document !== `undefined` and reuse it?

src/components/containers/portal.js
```diff
@@ -0,0 +1,53 @@
+import { Component } from 'react'
+import ReactDOM from 'react-dom'
+import PropTypes from 'prop-types'
+
+// Use a ternary operator to make sure that the document object is defined
```
OskarAhl: maybe the comment could explain why this check is needed?

src/components/containers/portal.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    componentWillUnmount = () => {
+        portalRoot.removeChild(this.el)
+    }
+
+    render() {
+        const { children, is_open } = this.props
+        if (gatsbyRoot) {
+            gatsbyRoot.style.filter = is_open ? 'blur(2px)' : 'none'
```
OskarAhl: can we pass the blur as a prop?

PR #15, opened 2019-07-03
Amam/signup

cakasuma:
**Setup**
- added favicons in `gatsby-plugin-manifest`
- added input reusable component
- added topography for text and header

**Interface**
- added signup interface
- added colors
- added icons


src/components/form/button.js
```diff
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types'
 
 const StyledButton = styled.button`
     border-radius: 6px;
+    padding: 1rem;
+    font-size: 100%;
```
arash-binary: Can we use `font-size: 1.6rem` instead of `font-size: 100%` ?

src/components/elements/signup-modal.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import styled from 'styled-components'
+import Row from '../containers/row'
+import Signup from '../form/signup'
+
+const SignupWrapper = styled.div`
```
OskarAhl: can be a `styled.article` since the modal is a "self-contained section"

src/components/elements/signup-modal.js
```diff
@@ -0,0 +1,60 @@
...
+    background-color: var(--color-grey-1);
+    width: 50vw;
```
OskarAhl: no max-width here?

src/components/elements/signup-modal.js
```diff
@@ -0,0 +1,60 @@
...
+    border-radius: 6px;
+`
+
+const LeftContent = styled.div`
```
OskarAhl: can left and right be combined? since they share a bunch of styling.

src/components/elements/signup-modal.js
```diff
@@ -0,0 +1,60 @@
...
+    max-width: 25rem;
+`
+const CaptionDesc = styled.p`
+    font-size: 1.6rem;
+    line-height: 1.5;
+    padding: 1.6rem;
+`
+
+const SignupModal = () => (
+    <SignupWrapper>
```
OskarAhl: think we can also add a `aria-hidden="true"` since it's only visible by default

src/components/form/signup.js
```diff
@@ -0,0 +1,145 @@
+import React, { Component } from 'react'
+import styled from 'styled-components'
+import Button from './button'
+import Input from './input'
+import Row from '../containers/row'
+import Facebook from 'images/svg/facebook.svg'
+import Google from 'images/svg/google.svg'
+
+const Title = styled.h2`
```
OskarAhl: Is this title style used at other pages? if so let's use a shared `<Header />`

src/components/form/input.js
```diff
@@ -1,10 +1,85 @@
...
+    left: 5px;
+    top: 10px;
+    transition: 0.25s ease all;
+`
+
+const Input = ({ label, ...props }) => (
+    <InputWrapper>
+        <StyledInput {...props} />
+        <Bar />
+        <StyledLabel>{label}</StyledLabel>
```
OskarAhl: the label needs a `for=..`

src/components/form/signup.js
```diff
@@ -0,0 +1,145 @@
...
+class Signup extends Component {
+    state = {
+        email: '',
+    }
+
+    handleInputChange = e => {
+        e.preventDefault()
+        const target = e.target
+        const value = target.value
+        const name = target.name
```
OskarAhl:
```suggestion
        const {name, value} = e.target
```

src/components/form/signup.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    handleLogin = () => {
+        // eslint-disable-next-line no-console
+        console.log('Hi login')
+    }
+
+    render() {
+        return (
+            <>
```
OskarAhl: SHould be inside a `<form />` with a submit

src/components/form/input.js
```diff
@@ -1,10 +1,85 @@
...
-const Input = ({ text }) => <input>{text}</input>
+const InputWrapper = styled.div`
+    position: relative;
+    width: 100%;
+`
+
+const StyledInput = styled.input`
+    background: none;
+    color: var(--color-black);
+    font-size: 1.6rem;
```
OskarAhl: we could consider adding a css variable for this since we are using this size everywhere also

src/components/elements/signup-modal.js
cakasuma: by the way we are using react, aria-hidden is unneeded

src/components/form/input.js
cakasuma: for needs an ID, we are doing it the react way to bind the text with the state, and for is optional actually, we can do it without adding too much id for each input

src/components/form/signup.js
cakasuma: handed to @arash-binary

src/components/elements/topography.js
```diff
@@ -0,0 +1,24 @@
...
+
+const createElement = React.createElement
+
+export const Text = styled.p`
+    font-weight: ${props => props.weight || 'normal'};
+    line-height: ${props => props.lh || '1.25'};
+    font-size: var(--text-size-s);
+`
+
+export const Header = styled(({ as = 'h1', children, ...props }) =>
```
OskarAhl: 👍 do we need color as a prop also?

src/pages/index.js
```diff
@@ -19,6 +20,26 @@ class Home extends Component {
         })
     }
 
+    handleEscape = e => {
```
OskarAhl: I think the keycodes should be handled in the modal component? you can pass down a function that sets `show_modal` to false

PR #25, opened 2019-07-12
Amam/tnc

cakasuma:
**addition**
- add reusable side tab component
- change animation for button
- improve speed for input animation
- add internal link gatsby for anchor tags
- add terms and conditions text


package.json
```diff
@@ -54,7 +54,7 @@
     "license": "Apache License 2.0",
     "scripts": {
         "build": "npm run test && gatsby build --prefix-paths",
-        "develop": "gatsby develop -p 80",
+        "develop": "sudo gatsby develop -p 80",
```
OskarAhl:
```suggestion
        "develop": "gatsby develop -p 80",
```

src/components/elements/tab.js
```diff
@@ -0,0 +1,54 @@
...
+    &.tab-active {
+        border-bottom: 1px solid var(--color-red);
+
+        & > p {
+            color: var(--color-red);
+        }
+    }
+`
+
+class Tab extends Component {
```
OskarAhl: Can change this to stateless component and move it to the `sideTab` file, since that is the only place where it's being used.

PR #35, opened 2019-07-19
arash/Deriv.com, sticky navbar + sliding in cta button + text changes

arash-binary:
Fixed Header:
    refactoring header component.
    adding animation to the header's button.
    modify Card's names in Card component.


src/components/layout/layout.js
```diff
@@ -1,14 +1,18 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import styled from 'styled-components'
 
 import Nav from './nav'
 import Footer from './footer'
 
 const Layout = ({ children }) => {
+    const Main = styled.main`
```
oskar-binary: can this be moved outside of `Layout`?

PR #56, opened 2019-07-25
arash/m-layout

arash-fs:
responsive: 

- navbar

- footer


src/components/layout/footer.js
```diff
@@ -117,61 +169,74 @@ const Social = styled.div`
...
+        p {
+            padding-top: 3rem;
+            padding: 3rem 0 0.7rem 0;
+        }
+        div {
+            display: flex;
+            justify-content: space-around;
+        }
+    }
+    // **********************************************
```
oskar-binary: can we remove this too and replace it with a `TODO` comment explaining why it should have display: none

arash-fs: Sure

PR #68, opened 2019-08-06
arash/render component based on device size

arash-fs:


src/components/containers/show.js
```diff
@@ -43,17 +44,21 @@ const Show = ({ children, to, device }) => {
         if (device) {
             switch (device) {
                 case 'laptop':
-                    //TODO: add breakpoints for laptop > window.innerwidth here, and update setInvisible memoized state
+                    setdeviceVisibility(
```
mustofa-binary: we can just use setVisible, combine them both, it will work the same

arash-fs: It's better to have separate states for device and region.
Sometimes we should consider rendering based on region, sometimes based on device.
So it's better to have different states

PR #80, opened 2019-08-07
arash/m-keepsafe

arash-fs:


src/components/elements/hero.js
```diff
@@ -12,14 +12,15 @@ const Wrapper = styled.section`
...
+    max-width: ${props => props.paragraph_max_width || ''};
 `
 
 const NoImageHero = styled(Container)`
     flex-direction: column;
     align-content: space-around;
     padding: 8rem 0 10rem 0;
 `
-const Hero = ({ header, paragraph, paragraph_width }) => (
+const Hero = ({ header, paragraph, paragraph_width, paragraph_max_width }) => (
```
mustofa-binary: is it possible to use either max_width or width instead of both

src/pages/keep-safe/index.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 import SEO from '../../components/containers/seo'
 import { SecurityIconGrid } from './_icon-grid'
+import Show from 'components/containers/show'
```
mustofa-binary: you can run npm run test, to fix all of import orders

src/components/elements/hero.js
arash-fs: yes, sometimes we need to set a limit on it.

src/pages/keep-safe/index.js
arash-fs: Sure

PR #98, opened 2019-08-19
amam/refactor_mobile

mustofa-binary:
use MediaQuery from react-responsive instead of window.innerwidth happens after rendering


src/components/containers/show.js
```diff
@@ -13,55 +12,35 @@ const handleEu = (setVisible, to) => is_eu_country => {
...
-                        window.innerWidth < size.tabletL.slice(0, -2),
-                    )
-                    break
-                default:
-                    break
-            }
+        const clients_country = Cookies.get('clients_country')
+        const showEu = handleEu(setVisible, to)
+        if (clients_country) {
+            isEuCountry(clients_country)
```
ashkanx:
```suggestion
          //  isEuCountry(clients_country)
```

mustofa-binary: we don't use show component anywhere at all as of now. we probably use this only if there is small difference between eu and non eu in a single page. otherwise we will remove the show component.

PR #101, opened 2019-08-19
arash/Hamburger menu

arash-fs:


src/images/svg/arrow.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
```
mustofa-binary: please add aria-labelledby and id

src/components/elements/off-canvas-menu.js
```diff
@@ -0,0 +1,90 @@
...
+
+const OffCanvasMenu = styled.section`
+    position: fixed;
+    background-color: var(--color-white);
+    top: 0;
+    height: 100vh;
+    width: 238px;
+    opacity: 0.98;
+    transition: right 0.4s;
+    right: ${props => (props.is_canvas_menu_open ? '0' : '-238px')};
```
mustofa-binary: can we use rem here?

arash-fs: Yes we can but since we have this off-canvas menu just in mobile, I prefer to use px instead of rem. Also, if we use rem it will be bigger in larger screen size devices, but we don't want to.
@oskar-binary what is your opinion?

oskar-binary: @arash-fs For me a fixed size for the mobile menu is fine here (px)

oskar-binary:
Could we add a `cursor:pointer` to the hamburger icon in this PR also?

arash-fs:
@oskar-binary Sure, I'll add `cursor: pointer` but in mobile, we haven't any cursor.

arash-fs:
@mustofa-binary Since the UI/UX haven't any specific design for tablet view it's better to use static width for the menu. we can change the width later if it's needed.

PR #118, opened 2019-08-22
amam/add_gclid

mustofa-binary:
# Description

Adding google click identifier in virtual account opening

Fixes # (issue)
- add gclid_url to req param

## Type of change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Dependency update (can be both non-breaking and breaking change)


src/components/form/signup.js
```diff
@@ -129,6 +129,7 @@ class Signup extends Component {
         const affiliate_token = Cookies.getJSON('affiliate_tracking')
         const signup_device = LocalStore.get('signup_device')
         const date_first_contact = LocalStore.get('date_first_contact')
+        const gclid = LocalStore.get('gclid')
```
negarn: where are you storing this? when will you remove it?

mustofa-binary: in traffic source, I set gclid based on url param

PR #120, opened 2019-08-22
Amam/improvements

mustofa-binary:
# Description

fix and improvements deriv com

Fixes # (issue)
- use Show and styled component for media query
- remove react-responsive
- update packages
## Type of change

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] Dependency update (can be both non-breaking and breaking change)
-   [x] This change requires a documentation update


src/components/containers/country.js
```diff
@@ -0,0 +1,53 @@
...
+            break
+        case 'non-eu':
+            setVisible(!is_eu_country)
+            break
+        default:
+            break
+    }
+}
+
+const Country = ({ children, to }) => {
```
oskar-binary:
```suggestion
const Location = ({ children, to }) => {
```

location is probably a better name since EU is not a country

PR #148, opened 2019-08-27
Playground netlify header cache

reza-fs:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

-   [*] Dependency update (can be both non-breaking and breaking change)

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -20,7 +20,6 @@ module.exports = {
         'gatsby-transformer-sharp',
         'gatsby-plugin-sharp',
         'gatsby-plugin-sitemap',
-        'gatsby-plugin-remove-serviceworker',
         {
             resolve: 'gatsby-plugin-manifest',
             options: {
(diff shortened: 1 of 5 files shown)


gatsby-config.js
```diff
@@ -20,7 +20,6 @@ module.exports = {
         'gatsby-transformer-sharp',
         'gatsby-plugin-sharp',
         'gatsby-plugin-sitemap',
-        'gatsby-plugin-remove-serviceworker',
```
oskar-binary: I think we still want to keep the empty service worker.

gatsby-config.js
```diff
@@ -93,5 +92,16 @@ module.exports = {
                 policy: [{ userAgent: '*', allow: '/' }],
             },
         },
+        {
+            resolve: `gatsby-plugin-netlify`,
+            options: {
+                headers: {
+                    '/page-data/*': [
+                        'cache-control: max-age=0,no-cache,no-store,must-revalidate',
```
oskar-binary: In the preview of this PR the `cache-control` seems to be as production? if it is the same why do we need this?

mustofa-binary: I think just `must-revalidate` should do to prevent cache invalidation, the cache is still needed for faster performance right? and limit max-age, probably like 1 week or 1 month

PR #155, opened 2019-08-29
Amam/enable translation

mustofa-binary:
# Description

Enable translation and add Content security policy

Changes:

-   enable translation

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [x] Breaking change
-   [ ] Dependency update


oskar-binary:
can you show the language dropdown for non-production also? + send the selected language to the socket connection in (socket_base)

mustofa-binary:
> can you show the language dropdown for non-production also? + send the selected language to the socket connection in (socket_base)

we dont have design for language dropdown yet, sure will send to socket base

src/common/utility.js
```diff
@@ -38,6 +38,10 @@ const getPropertyValue = (obj, k) => {
     return obj ? cloneObject(obj[keys[0]]) : undefined
 }
 const getLocationHash = () => (location.hash ? location.hash.substring(1) : '')
+
+const getLanguage = () =>
+    typeof localStorage !== 'undefined' ? localStorage.getItem('i18n') : null
```
oskar-binary: should return `en` as default I think if there is no `i18n` in localStorage

src/pages/ach.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
```
oskar-binary: I don't think `ach` should be a separate page 🤔 I think it's enough to enable in-context translation if we change the current language but keep the base same as english `deriv.com` if that is possible

i18n-config.js
```diff
@@ -6,48 +6,48 @@ module.exports = {
...
+        path: 'it',
+        display_name: 'Italiano',
+    },
+    pl: {
+        path: 'pl',
+        display_name: 'Polish',
+    },
+    de: {
+        path: 'de',
+        display_name: 'Português',
```
oskar-binary: `de` should be German

oskar-binary:
deployment is failing

i18n-config.js
```diff
@@ -6,48 +6,56 @@ module.exports = {
         display_name: 'English',
     },
     // TODO: uncomment this when translations are ready
```
oskar-binary: we can remove this comment now also

src/components/localization/config.js
```diff
@@ -8,19 +8,20 @@ i18n.use(initReactI18next).init({
     // TODO: Optimize translation file loading with GraphQL
     // TODO: Enable this when translations are ready
```
oskar-binary: and these 2

PR #158, opened 2019-09-01
arash/deriv-bugs-0

arash-fs:
# Description

Changes:
- Markets number
- Slider bullet
- Footer logo navigation
- Trade logo tooltips

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [ ] Dependency update


src/components/elements/carousel.js
```diff
@@ -28,12 +28,12 @@ const Bullet = styled.button`
     width: 1rem;
     height: 1rem;
     border-radius: 1rem;
+    border: none;
```
mustofa-binary: how does this affect?, i think the issue is fixed in staging

arash-fs: Safari and Edge automatically add shadow to the button's border :D 
So if we haven't any border, we haven't any shadow.
Also, this line fixes the chrome issue when you push mouse down but not release it. 
in this case, you see a half border on the buttons
So it's better to have this one

mustofa-binary: fyi, `border: none` already exists on line 36, changing it to line 31 which is in the same specificity will not make any difference. and if its border: none, we should also delete line 30 border-radius: 1rem, since it wont make any effect.

PR #194, opened 2019-09-20
amam/reset_password [WIP]

mustofa-binary:
# Description

add reset password flow

Changes:

- add '/reset-password' page
- add nav-static and layout-static component
- call verify email for reset password
- add '/choose-password' page
- call reset password api

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Breaking change
-   [ ] Dependency update


gatsby-browser.js
```diff
@@ -6,11 +6,18 @@ import isMobile from './src/common/os-detect'
 import './src/components/localization/config'
 import 'typeface-ibm-plex-sans'
 
+const browser = typeof window !== 'undefined' && window
```
oskar-binary:
```suggestion
const is_browser = typeof window !== 'undefined' && window
```

PR #202, opened 2019-09-25
arash/DTrade static

arash-fs:
# Description

Summary of the change and/or issue that is fixed.

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] This change requires a documentation update

PR #205, opened 2019-09-26
amam|kelly/dbot_page

kelly-binary:


arash-fs:
- in `how it works` section we should render a video there, I'm working on it in this https://github.com/binary-com/deriv-com/pull/202.
- in `Get integrated help` the margins is not the same as the Zeplin.
- `go to live demo` color does not match the Zeplin.

arash-fs:
@mustofa-binary BTW, I create a component which handles the video

PR #214, opened 2019-09-30
arash/Newdmt5

arash-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [ ] Dependency update


src/pages/dmt5/_dtrader-video.js
```diff
@@ -0,0 +1,282 @@
...
+`
+const VideoWrapper = styled.div`
+    grid-area: video;
+    margin-top: 4rem;
+    position: relative;
+    text-align: center;
+    display: flex;
+    justify-content: center;
+`
+// const MacbookFrame = styled(MacBook)`
```
oskar-binary: Why all this commented out code? if it's not needed please remove it as we can always retrieve it from our source control if we need it in the future.

PR #230, opened 2019-10-02
Sara / Fixdesign

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [ ] Dependency update


arash-fs:
The cards are not aligned in the mobile view.
you should define a different template for mobile from the desktop in css grid

arash-fs:
@sara-fs Please check the titles

src/components/layout/footer.js
```diff
@@ -165,8 +165,8 @@ const RiskNote = styled.section`
     }
 `
 const FooterStyledLink = styled(StyledLink)`
+    font-size: ${props => (props.edgeIssue ? '1.39rem' : '1.4rem')};
```
mustofa-binary: please use text-size from variables.js as fallback

PR #246, opened 2019-10-07
Oskar/home page changes

oskar-binary:
* New homepage design

* New Carousel

* Ticker: should show 7 forex 3 volatilities - if forex is closed show 7 volatilities 3 forex where forex has a static value of last open value


src/components/elements/carousel.js
```diff
@@ -1,88 +1,230 @@
...
-    & > svg {
-        position: absolute;
-        right: 2rem;
-        bottom: 2rem;
+    div {
+        display: flex;
+        align-items: center;
+        margin-top: 3rem;
+        width: 100%;
+
```
mustofa-binary:
```suggestion
            /* TODO: use <Text /> component instead */
```

PR #274, opened 2019-10-16
amam/fix_localize

mustofa-binary:
# Description

- fix tricky strings cant be translated

Changes:

- use localize inside of the one called the array in help centre and carousel component
- dont go back to homepage after switch language
-- add non greedy regex to match currentpath to the new language path

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [ ] Dependency update


src/components/layout/nav.js
```diff
@@ -199,18 +197,19 @@ const Nav = () => {
...
-                        </NavRightContainer>
+                        <Button onClick={handleLogin} primary>
+                            <span>{localize('Log in')}</span>
+                        </Button>
+                        <SignupButton
+                            secondary
+                            onClick={toggleModal}
+                            move={show_button}
+                            ref={button_ref}
+                            width={button_ref}
```
oskar-binary: We only need to pass the ref once and then use the ref inside signupButton to get the width

src/components/localization/language-switcher.js
```diff
@@ -31,12 +31,36 @@ class LanguageSwitch extends Component {
...
+                can be something like /es/about/
+                or just /about/
+                or just /
+            
+            */
+            const current_path = window.location.pathname
+            let destination_path = path
+
+            const current_lang = localStorage.getItem('i18n') || 'en'
+            if (current_lang === 'en') {
```
oskar-binary:
```suggestion
            if (value === '/en') {
```

mustofa-binary: value is the destination path, meanwhile, we want to know the what is the current url path

oskar-binary: line 47 value={`/${this.state.language}`}

PR #301, opened 2019-10-24
arash/ 1 request for auto carousel

arash-fs:
# Description

reduce number of send requests to the server from 10 to 1 on weekdays and 4 in holidays.
in holidays we show 3 close symbols, in that case, we should use `tick history` API and in `tick history` we can't send an array likes `tick stream` so we need to send 3 requests for close symbols and 1 for open symbols.

Changes:

- separate tick from ticker
- add an Event Emitter for handling request in the auto carousel
- fix PropTypes warning in Card.js

## Type of change

-   [x] Bugfix
-   [x] New feature


src/pages/home/_ticker.js
```diff
@@ -194,21 +140,69 @@ const getTickerMarkets = active_symbols => {
     })
     if (volidx.length) volidx = shuffle(volidx).slice(0, volatility_count)
     if (forex.length) forex = shuffle(forex).slice(0, forex_count)
-
-    return [...volidx, ...forex]
+    ;[...volidx, ...forex].forEach(symbol =>
```
oskar-binary:
```suggestion
    [...volidx, ...forex].forEach(symbol =>
```

src/pages/home/_ticker.js
```diff
@@ -222,23 +216,31 @@ class Ticker extends React.Component {
             forget_all: 'ticks',
         })
     }
+    shouldComponentUpdate() {
+        if (this.count < 20) {
```
oskar-binary: what is this count for?

src/pages/home/_event-emitter.js
```diff
@@ -0,0 +1,16 @@
+const EventEmitter = {
```
oskar-binary: 👍

mustofa-binary:
@arash-fs conflict

src/pages/home/_ticker.js
arash-fs: this is the prettier suggestion.

src/pages/home/_ticker.js
```diff
@@ -189,21 +44,66 @@ const getTickerMarkets = active_symbols => {
         synthetic = shuffle(synthetic).slice(0, volatility_count)
     }
     if (forex.length) forex = shuffle(forex).slice(0, forex_count)
-
-    return [...synthetic, ...forex]
+    ;[...synthetic, ...forex].forEach(symbol =>
```
oskar-binary: still has a `;` here that shouldn't be here, if it's from prettier you can add a prettier ignore to this line else put it in a variable

arash-fs: done

PR #349, opened 2019-11-20
oskar/add_lottie

oskar-binary:
# Description
* Add react-lottie library
* Add component wrapper for Lottie

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/components/elements/lottie.js
```diff
@@ -0,0 +1,44 @@
...
+            preserveAspectRatio: 'xMidYMid slice',
+        },
+    }
+
+    const eventListeners = []
+    if (onAnimationComplete) {
+        eventListeners.push({ eventName: 'complete', callback: onAnimationComplete })
+    }
+
+    return (
```
mustofa-binary: can we add example in the implementation-guide.md, and add this https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio in it 🙏

PR #365, opened 2019-12-03
arash/Dtrader

arash-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/dtrader/_dtrader-video.js
```diff
@@ -41,24 +44,25 @@ const StepCommon = css`
...
+            : 'opacity: 0.2; border-left: 5px solid rgb(0, 0, 0, 0)'};
 `
 const Step3 = styled(Header)`
     ${StepCommon}
     ${props =>
         props.current_time >= 13
-            ? 'color: var(--color-red); border-bottom: 1px solid var(--color-red)'
-            : 'color: var(--color-red-2); border-bottom: 1px solid var(--color-red-2)'};
+            ? 'color: var(--color-black-3); border-left: 5px solid var(--color-red)'
+            : 'opacity: 0.2; border-left: 5px solid rgb(0, 0, 0, 0)'};
```
sara-fs: Could we have one Step component instead of all those three?
```

const Step = styled(Header)`
    ${StepCommon}
    margin-top: 0;
    ${props =>
             props.current_time >= props.minimun_time && props.current_time < props.maximum_time
            ? 'color: var(--color-black-3); border-left: 5px solid var(--color-red)'
            : 'opacity: 0.2; border-left: 5px solid rgb(0, 0, 0, 0)'};
`

```

arash-fs: They should be separated components because we have at least one true condition and the result will be true and all the steps return true declared value.

sara-fs: In my point of view, it will not affect the result. You will have one component but you can call them 3 times with different props. I think in this piece of code you are just deciding about the style that can be written once. :)

arash-fs: Aha, now I get it,
Sure, I will fix it :pray:

PR #366, opened 2019-12-03
arash/Dbot

arash-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/dbot/_dbot-video.js
```diff
@@ -0,0 +1,159 @@
+import React from 'react'
```
oskar-binary: Can the <Step />  be a shared component? then we don't have to repeat the styling and the logic for the places where we want to use it + it will be needed for other pages as well.

oskar-binary: By `this` I mean the Tabs Stepper

arash-fs: I agree it can be a shared component but, since I work on a separated branch for each page.
we merge them together after the pages pass QA.

PR #377, opened 2019-12-17
Retrieved the old code for show.js

sara-fs:
# Description

The old code for `show.js` was retrieved due to some problems which had been arised by `react-responsive`. Some styles were misapplied on production build (not on gatsby build) when using `react-responsive`. The problem was observed on reloading (refreshing) some pages such as `About Page` and `DMT5`.

Moreover, searhing through some forums, `react-media` has also the same problem. 

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


oskar-binary:
@sara-fs please write a quick summary of the issue in the description

PR #388, opened 2019-12-29
Created shared linkbutton

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/components/form/linkButton.js
```diff
@@ -0,0 +1,28 @@
...
+
+    &:hover {
+        background-color: var(--color-red-3);
+        border-color: var(--color-red-3);
+    }
+    &:focus,
+    &:active {
+        outline: none;
+    }
+    &:active {
```
mustofa-binary: why there are two `:active` styling, can combine into one

src/components/form/linkButton.js
```diff
@@ -0,0 +1,26 @@
+import styled from 'styled-components'
+import { LocalizedLink } from 'components/localization'
+
+const LinkButton = styled(LocalizedLink)`
```
oskar-binary: LinkButtons can also be `primary, secondary, tertiary`? i'd suggest moving the component inside button.js and share the css between button and LinkButton

PR #389, opened 2019-12-30
amam/utm_last_attribution_model

mustofa-binary:
# Description

make the utm follows last attribution model, which will make the data overridable.

Changes:

-  remove the truth checking on utm `setData` function

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/common/traffic-source.js
```diff
@@ -60,8 +59,8 @@ const TrafficSource = (() => {
 
         if (params.utm_source) {
             // url params can be stored only if utm_source is available
-            param_keys.map(key => {
-                if (params[key] && !current_values[key]) {
+            param_keys.forEach(key => {
```
mustofa-binary: foreach is faster than map, and also this is not a returning value loops

src/common/traffic-source.js
```diff
@@ -4,14 +4,13 @@ import { CookieStorage, LocalStore } from './storage'
...
-        if (!cookie) {
-            cookie = new CookieStorage('utm_data')
-            cookie.read()
-            // expiration date is used when writing cookie
-            const now = new Date()
-            cookie.expires = now.setMonth(now.getMonth() + 3)
-        }
+    const initUtmCookie = () => {
+        cookie = new CookieStorage('utm_data')
+        cookie.remove()
```
oskar-binary: we need to remove here even though it's declared as a `new CookieStorage` on the line before?

mustofa-binary: to select the `utm_data` cookie by `storage` service, we need to do `new CookieStorage('utm_data')`, then we can remove the previous data. but in this case, we dont need `cookie.read()` anymore

PR #413, opened 2020-01-10
oskar/new_landing_sections

oskar-binary:
# Description

* Simple steps cards
* Client testimonials carousel
* Signup banner

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/home/_markets.js
```diff
@@ -36,12 +35,12 @@ const MarketCard = styled.article`
     }
 `
 const Markets = () => (
-    <SectionContainer background="grey-10">
+    <SectionContainer style={{ backgroundColor: 'rgba(242, 243, 244, 0.3)' }}>
```
arash-fs: do we allow to use inline styles?

oskar-binary: @arash-fs I personally don't mind inline styling if it's for adding one or two properties but it could be changed to a styled component too

PR #430, opened 2020-01-16
Transferred About us page to new branch

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/images/svg/cyberjaya-pin-location.svg
```diff
@@ -0,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="204" height="53" viewBox="0 0 204 53">
```
oskar-binary: aria label and title to all svgs please 🙏

src/pages/about/_our-offices.js
```diff
@@ -0,0 +1,209 @@
...
+
+                <MapLink
+                    rel="noopener noreferrer"
+                    target="_blank"
+                    href="https://maps.google.com/?q=Lake+Plaza+Dubai&25.079220,55.149569&ll=25.079220,55.149569&z=16"
+                >
+                    <Pinpoint top="30.6%" left="63.7%">
+                        <Oval />
+                        <DubaiWrapper />
+                        {/* <LabelWrapper>{localize('Dubai')}</LabelWrapper> */}
```
oskar-binary: can we add a comment as to why the <LabelWrappers /> are commented?

src/pages/about/_join-us.js
```diff
@@ -0,0 +1,63 @@
...
+import { localize, LocalizedLink } from 'components/localization'
+
+const JoinUsContainer = styled.section`
+    display: flex;
+    justify-content: flex-end;
+    width: 100%;
+    padding: 4rem 9.7rem;
+    margin: auto;
+    align-items: right;
+    background-image: linear-gradient(to left, #232232, #223122 81%, rgba(0, 0, 0, 0.26) 50%);
```
mustofa-binary: we can use rgba for this instead of hex

sara-fs: This section was removed. Probably they will change it. I'll consider it in new implementation. Thanks.

PR #486, opened 2020-02-19
Sara / Safari nav issue

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release

### DIFF
--- a/src/components/layout/nav.js
+++ b/src/components/layout/nav.js
@@ -89,16 +89,16 @@ const NavRight = styled.div`
     }};
     transform: translateX(
         ${props => {
-        if (props.move) {
-            return 0
-        } else {
-            if (props.button_ref.current && props.mounted) {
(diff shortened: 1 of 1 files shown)


src/components/layout/nav.js
```diff
@@ -89,16 +90,16 @@ const NavRight = styled.div`
     }};
     transform: translateX(
         ${props => {
-        if (props.move) {
-            return 0
-        } else {
-            if (props.button_ref.current && props.mounted) {
-                const calculation = props.button_ref.current.offsetWidth + 2
-                return `${calculation}px`
+            if (props.move) {
```
mustofa-binary: please check if we can fix this in useEffect (component did update)

PR #490, opened 2020-02-20
chaks/circle-docker: added circleci pipeline with Dockerfile and nginx config

shudarshon-binary:
# Description

Changes:

-   added circle ci pipeline
-   added Dockerfile
-   added nginx configuration 

## Type of change

-   [ ] Bug fix
-   [X] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


deriv.com.conf
```diff
@@ -0,0 +1,55 @@
+map $http_cf_ipcountry $redirected_url {
```
oskar-binary: @shudarshon-binary the `/eu` page is not ready yet so we can save this for later 🙏

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+          filters:
+            branches:
+              only: /^dev/
+  build_and_push_master:
+    jobs:
+      - release:
+          filters:
+            branches:
+              only: /^master/
+          context: binary-frontend-artifact-upload  # using context to retrieve shared secret from circle ci
```
mahdipourziaei-binary: Can we add staging release to merge with `dev` too?

deriv.com.conf
```diff
@@ -0,0 +1,55 @@
...
+    location ~ /\.git {
+        return 404;
+    }
+
+    location / {
+        if ($http_cf_ipcountry) {
+            return 301 https://$host$redirected_url$request_uri;
+        }
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
```
mahdipourziaei-binary: We'd need to use the request path for resolution of the html files here too. Currently this would serve `index.html` for all, right?

.circleci/config.yml
shudarshon-binary: yeah, sure. i will be writing the pipeline as well

deriv.com.conf
shudarshon-binary: i'll edit the nginx config

.circleci/config.yml
```diff
@@ -0,0 +1,177 @@
...
+          command: npm run build
+  compress:
+    description: "Compress"
+    steps:
+      - run:
+          name: "Compress"
+          command: |
+            pushd public/
+            tar -cvf artifact.tar *
+            mv artifact.tar ${OLDPWD}/
```
4p00rv: What's `OLDPWD` containing? Where do we get this value from?

shudarshon-binary: its a unix variable which means old working directory. once you cd into different directory then you can echo ${OLDPWD} to get your previous directory

.circleci/config.yml
```diff
@@ -0,0 +1,178 @@
...
+        environment:
+          - GATSBY_CPU_COUNT=2
+    steps:
+      - git_checkout_from_cache
+      - npm_install_from_cache
+      - build
+      - compress
+      - upload_artifact  # uploading the built code to s3 to create a backup of key services separate from Kubernetes deployment
+      - upload_checksum  # uploading compressed artifact checksum to cross match artifact fingerprint
+      - docker_build
```
4p00rv: This would overwrite latest images for staging and production, we should distinguish the images for staging and production. I suggest you have a separate tag for staging as latest-staging

shudarshon-binary: got it! changing as per new requirement

PR #493, opened 2020-02-21
Negar-Sara/update help centre

negarn:
# Description

Add the new help centre articles

Changes:

- New content added for account
- New content added for trading
- New content added for deposit and withdrawals
- New content added for DMT5
- New content added for DTrader
- New content added for DBot
- New content added for Security

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/home/_hero.js
```diff
@@ -182,7 +182,7 @@ export const Hero = () => {
                             </CheckMarkBullet>
                         </CheckMarkList>
                         <ButtonWrapper jc="unset">
-                            <HeroButton type="submit" secondary="true" to="/signup/">
+                            <HeroButton type="submit" secondary to="/signup/">
```
sara-fs: In the latest code reviews I was told to add the value for syled-component: secondary="true".
Which one is correct? @negarn @oskar-binary

src/pages/help-centre/_help-articles.js
```diff
@@ -2,95 +2,391 @@ import React from 'react'
...
+                },
+                {
+                    title: <Localize translate_text="Why were my documents declined?" />,
+                    category: 'Security',
+                    sub_category: <Localize translate_text="Verification" />,
+                    label: 'documents-declined',
+                },
+            ],
+        },
+    ],
```
sara-fs: Is there any necessity to use an `array of array` at the first level? The search is not working anymore and I think it's due to the `match-sorter` package which is used in the code; Apparently, it only can handle a flat array. I think we have to either change the function of `getAllArticles()` or `_help-articles` array. What's your idea?

src/pages/home/_hero.js
negarn: I see, in deriv-app we have the styleguide to not add true values to boolean parameters as they will be true by default, so I was following that. please keep the true then.

src/pages/help-centre/_help-articles.js
negarn: I did it so that it could show three items in one row. I think it makes it confusing to use this structure to pass to DOM, please update it however you see fit. If we create components I think it's more readable than an array, but you know the codebase better than me.

src/pages/home/_hero.js
oskar-binary: @negar-binary for react components yes but not for styled-components, will check if there is any way to get around the warning message.

src/pages/test/index.js
```diff
@@ -0,0 +1,240 @@
...
+        return (
+            <Layout>
+                <StyledDiv>
+                    {splitted.map((article, id) => {
+                        // console.log('article', article)
+                        return (
+                            <RowDiv key={id}>
+                                {article.map((item, idx) => {
+                                    {
+                                        // console.log('item', item.articles)
```
mustofa-binary: what is this page used for?

mustofa-binary: please delete this page

src/components/custom/_dhero.js
```diff
@@ -128,14 +128,14 @@ const DHero = ({
                     <div>
                         <LinkWrapper>
                             {join_us_for_free ? (
-                                <TryForFree secondary>
+                                <TryForFree secondary="true">
```
mustofa-binary: no need to pass in `="true"`, just secondary is enough since its expecting boolean

src/components/custom/_dhero.js
```diff
@@ -128,14 +128,14 @@ const DHero = ({
...
                                     <LocalizedLink to="/signup/">
                                         {localize('Create demo account')}
                                     </LocalizedLink>
                                 </TryForFree>
                             ) : null}
                             {start_automating ? (
-                                <TryForFree secondary>
+                                <TryForFree secondary="true">
```
mustofa-binary: same for all of the other buttons

src/pages/help-centre/account.js
```diff
@@ -25,199 +25,178 @@ export const StyledLink = css`
...
+        </Header>
+        <Text>
             <Localize
-                translate_text="You have read the <0>Legal Terms and Conditions</0> in full."
+                translate_text="If you have not made a deposit or created a Deriv MT5 (DMT5) account, you can change your name, date of birth, or citizenship by going to <0>Settings ></0> <1>Personal details</1>."
                 components={[
+                    <strong key={0} />,
                     <ExternalLink
-                        href={url}
+                        href="https://deriv.app/account/personal-details"
```
mustofa-binary: you can use deriv app url on common/utility

src/components/custom/_dhero.js
negarn: but @oskar-binary said to include it to avoid the warning in console :D so which one is right?

mustofa-binary: well.. in that case its okay, can skip it 👍 the solution for those console error is here https://styled-components.com/docs/faqs#why-am-i-getting-html-attribute-warnings . adding ="true" wont be avoid all the warning console actually. we will do that in separate card instead

src/pages/help-centre/_search-results.js
```diff
@@ -75,6 +75,11 @@ const StyledLink = styled(Link)`
     text-decoration: none;
     color: black;
     font-size: var(--text-size-s);
+
+    :hover {
+        color: red;
```
oskar-binary:
```suggestion
        color: var(--color-red);
```

PR #511, opened 2020-02-27
oskar/update_signup_page

oskar-binary:
# Description
* Update text
* Add login redirect to notebox


src/components/custom/signup.js
```diff
@@ -91,6 +91,7 @@ class Signup extends Component {
         this.setState({
             [name]: value,
         })
+        this.handleValidation(value)
```
mustofa-binary: can you help to make a todo to change this to formik 🙏

oskar-binary: Will add a trello card to the backlog

PR #513, opened 2020-02-27
Sara / Split about us on footer

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/about/index.js
```diff
@@ -22,20 +121,102 @@ const About = () => {
...
+                                            'Jean-Yves has been an entrepreneur since the age of 20. From 1997 to 1999, he developed the algorithms that would become one of the world’s first trading platforms. He was granted a patent for his binary options trading system in 2007, and granted two more patents in 2011 for systems and methods that enable financial market speculation.',
+                                        )}
+                                    </Text>
+                                </div>
+                            </LeadershipWrapper>
+                        </ContentWrapper>
+                    )}
+                </StyledContainer>
+            </Background>
+            {active_tab === 'story' && <OurStory />}
```
mustofa-binary: isnt this the same as above, we can combine them right?

src/pages/about/index.js
```diff
@@ -1,19 +1,118 @@
...
+    return [active_tab, setTab]
+}
+const About = () => {
+    const [active_tab, setTab] = useTabState('story')
+    const is_story = active_tab === 'story'
+    const is_leadership = active_tab === 'leadership'
+    useEffect(() => {
+        const new_tab = getLocationHash() || 'story'
+        setTab(new_tab)
+    })
```
oskar-binary: I think we only want to this to fire on mount?

```suggestion
    }, [])
```

oskar-binary: oops didn't mean to approve it!

src/pages/about/index.js
sara-fs: Oh yes I changed it to: 
```
{is_story && <OurStory />}
            {is_leadership && <Leaders />}
```

src/pages/about/index.js
oskar-binary: actually nevermind, you can change it back to the old way as in this way the component is not re-rendered when you update the locationhash from the link in the footer.

PR #565, opened 2020-03-24
oskar/datadog_RUM

oskar-binary:
# Description

* load data dog RUM script in <head> for production

### DIFF
--- a/src/components/containers/seo.js
+++ b/src/components/containers/seo.js
@@ -154,8 +154,8 @@ const SEO = ({ description, meta, title, no_index }) => {
             <script>
                 {`window.location.hostname === 'www.deriv.com' && window.DD_RUM &&
                     window.DD_RUM.init({
-                        clientToken: 'pubc006d593980ec5cd3dcf299ba19a9335',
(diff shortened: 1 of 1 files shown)


src/components/containers/seo.js
```diff
@@ -146,7 +146,19 @@ const SEO = ({ description, meta, title, no_index }) => {
                       ]
                     : []),
             ].concat(meta)}
-        ></Helmet>
+        >
+            <script
```
mustofa-binary: so we will still load datadog js in localhost?

oskar-binary: @mustofa-binary loading the the script yes, but it should only run in production from the `window.location.hostname === 'www.deriv.com' &&...` check

mustofa-binary: is it possible to defer?

oskar-binary: I think we don't want to defer since data dog will measure load time start and end times

mustofa-binary: ah, i see, alright then 👍

PR #607, opened 2020-04-09
oskar/update_dd_check

oskar-binary:
# Description
Update data dog initialization check

### DIFF
--- a/src/components/containers/seo.js
+++ b/src/components/containers/seo.js
@@ -154,7 +154,7 @@ const SEO = ({ description, meta, title, no_index }) => {
                 type="text/javascript"
             ></script>
             <script>
-                {`window.location.hostname === 'www.deriv.com' && window.DD_RUM &&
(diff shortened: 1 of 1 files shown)


src/components/containers/seo.js
```diff
@@ -154,7 +154,7 @@ const SEO = ({ description, meta, title, no_index }) => {
                 type="text/javascript"
             ></script>
             <script>
-                {`window.location.hostname === 'www.deriv.com' && window.DD_RUM &&
+                {`(window.location.hostname === 'www.deriv.com' || window.location.hostname === 'deriv.com') && window.DD_RUM &&
```
mustofa-binary, line 157: we dont use www.deriv.com hostname anymore, we resolved the hostname to always use `deriv.com`

oskar-binary: ok so we just need window.location.hostname === 'deriv.com' then?

PR #762, opened 2020-05-19
Sara / Removed hyper link from navbar

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/components/layout/nav.js
```diff
@@ -178,6 +178,12 @@ const NavLink = styled.li`
 `
 const StyledLink = styled(LocalizedLink)`
     ${SharedLinkStyle}
+
+    &:hover {
```
mustofa-binary, line 182: we only want to remove red line on logo, not the entire navbar. tips: use localized link instead of styled link for logo on reset password

PR #784, opened 2020-05-28
Sara / Made the home hero responsive

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/pages/home/_hero.js
```diff
@@ -104,8 +108,8 @@ const ButtonWrapper = styled(Flex)`
 `
 
 const StyledVideo = styled.video`
-    position: absolute;
     opacity: 0.5;
+    max-height: 82.7rem;
```
mustofa-binary, line 112: the container for hero need to be the same as well, because currently afaik, its 87vh, so there will be blackened background at the bottom of the video 🙏

sara-fs: Actually it was due to an extra padding on `herowrapper` on laptop screen. I removed it and it's ok now. When I add  `87vh` the problem still exists.

PR #794, opened 2020-05-31
Sara / Updated signup new design

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/components/custom/_signup-new.js
```diff
@@ -37,21 +38,23 @@ const Line = styled.div`
...
     width: 100%;
-    padding: 1.6rem 1.6rem;
+    height: 7.7rem;
+    padding: 0.8rem 0.8rem;
     margin-top: 1.6rem;
-    border-radius: 0.6rem;
-    background-color: var(--color-grey-8);
-    border: solid 1px var(--color-grey-17);
+    border-radius: 0.4rem;
+    background-color: rgba(242, 243, 244, 0.56);
```
arash-fs, line 57:
```suggestion
    background-color: var(--color-grey-8);
    opacity: 0.56;
```

src/components/elements/checkbox.js
```diff
@@ -35,7 +35,6 @@ const StyledCheckbox = styled.div`
         if (props.secondary && props.checked) return '2px solid var(--color-red)'
         return '2px solid var(--color-grey-5)'
     }};
-    border-radius: 3px;
```
arash-fs, line 38: we are using this element over the website, would you please change the `border-radius` inside the inheritance component if it's needed?

src/components/custom/_signup-new.js
sara-fs: I don't know why, but the result is not the same.

src/components/elements/checkbox.js
sara-fs: `border-radius` has been removed from checkbox style throughout the project.

PR #798, opened 2020-06-01
oskar/gdpr_cookie_consent

oskar-binary:
- (new) Cookie accept and decline for EU with new datalayer push event
- (fix) `Show.EU` and non-eu components with react context


src/components/layout/layout.js
```diff
@@ -16,22 +17,77 @@ const Main = styled.main`
...
+            window.addEventListener('load', windowLoaded)
         }
+
+        return () => window.removeEventListener('load', windowLoaded)
     }, [])
+
+    React.useEffect(() => {
+        if (!clients_country || !has_window_loaded) return
+
+        const is_eu_country = isEuCountry(clients_country)
```
mustofa-binary, line 65: means this will be always called on every page load then?

src/components/containers/show.js
```diff
@@ -81,18 +27,26 @@ export const Mobile = ({ children, ...props }) => (
...
+    const { is_eu } = React.useContext(LocationContext)
+
+    if (is_eu) return <>{children}</>
+    else return null
+}
+
+export const NonEU = ({ children }) => {
+    const { is_eu } = React.useContext(LocationContext)
+
+    if (!is_eu && typeof is_eu === 'boolean') return <>{children}</>
```
mustofa-binary, line 40: is there a case where its not a boolean?

mustofa-binary, line 40:
```suggestion
    if (is_eu === false) return <>{children}</>
```

src/components/form/button.js
```diff
@@ -73,7 +73,7 @@ const SharedButtonStyle = css`
             `
         if (props.white)
             return css`
-                border: 2px solid var(--color-white);
+                border: 2px solid var(--color-black-5);
```
mustofa-binary, line 76: any reason why?

src/components/layout/footer.js
```diff
@@ -281,319 +275,320 @@ const Footer = ({ has_banner_cookie }) => (
                                 </ExternalLink>
                             </div>
                         </SocialWrapper> */}
-                        {/* <div>{!no_language && <LanguageSwitcher />}</div> */}
-                        {/* <div>
+                            {/* <div>{!no_language && <LanguageSwitcher />}</div> */}
```
mustofa-binary, line 278: pls remove this 🙏

src/components/layout/layout.js
oskar-binary: it will be called once when window has load and the client has a country  - (changing route in gatsby is not counted as a window load event)

src/components/layout/layout.js
```diff
@@ -16,22 +17,77 @@ const Main = styled.main`
     position: relative;
 `
 
+const is_browser = typeof window !== 'undefined'
```
mustofa-binary, line 20: import { isBrowser } from 'common/utility'

src/components/form/button.js
oskar-binary: It's from Zeplin - and as far as I can see this type of button is not used anywhere else so should be ok?

src/components/layout/layout.js
oskar-binary: @mustofa-binary moved has_window_loaded to our global context instead

PR #910, opened 2020-06-29
Pull based release procedure with A/B testing

gokula-krishna-dev:


gokula-krishna-dev:
> AFAIK, [GATSBY](https://github.com/binary-com/deriv-com/blob/d9b7b3a52d9e0435cc83861d8af9bd1f0e325f97/.circleci/config.yml#L141) is no longer used now. All GATSBY environment variables can be removed. One confirmation from FE team would still be nice.

They still use Gatsby

gokula-krishna-dev:
> I think `notify_slack` configuration block should not be present multiple times. Project name and version for slack notification can be parameterized for using dynamic values. [this](https://github.com/binary-com/deriv-com/blob/d9b7b3a52d9e0435cc83861d8af9bd1f0e325f97/.circleci/config.yml#L128) one can be removed

Removed

PR #939, opened 2020-07-08
amam/livechat_integration

mustofa-binary:
# Description

New livechat integration

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


src/html.js
```diff
@@ -0,0 +1,41 @@
...
+                    name="viewport"
+                    content="width=device-width, initial-scale=1, shrink-to-fit=no"
+                />
+                {props.headComponents}
+            </head>
+            <body {...props.bodyAttributes}>
+                {props.preBodyComponents}
+                <div key={`body`} id="___gatsby" dangerouslySetInnerHTML={{ __html: props.body }} />
+                {props.postBodyComponents}
+                <script
```
oskar-binary, line 20: Not adding it with GTM anymore?

src/components/layout/layout.js
```diff
@@ -121,14 +144,35 @@ const Layout = ({ children, type, interim_type, padding_top, no_login_signup })
...
+                            typeof LC_API.chat_window_hidden() === 'undefined'
+                                ? true
+                                : LC_API.chat_window_hidden()
+                        if (is_chat_hidden) {
+                            LC_API.open_chat_window()
+                        } else {
+                            LC_API.hide_chat_window()
+                        }
+
+                        return false
```
oskar-binary, line 167: return is not needed here?

src/components/layout/layout.js
```diff
@@ -121,14 +144,35 @@ const Layout = ({ children, type, interim_type, padding_top, no_login_signup })
...
                     onAccept={onAccept}
                     onDecline={onDecline}
                     is_open={show_cookie_banner}
                 />
             )}
+            {is_livechat_interactive && LC_API && (
+                <LiveChat
+                    onClick={() => {
+                        const is_chat_hidden =
+                            typeof LC_API.chat_window_hidden() === 'undefined'
```
oskar-binary, line 158:
```suggestion
                            typeof LC_API.chat_window_hidden === 'undefined'
```

src/components/layout/layout.js
```diff
@@ -56,6 +73,12 @@ const Layout = ({ children, type, interim_type, padding_top, no_login_signup })
         }
     }, [])
 
+    React.useEffect(() => {
+        LC_API.on_after_load = () => {
+            setLiveChatInteractive(true)
+        }
+    }, [LC_API])
```
oskar-binary, line 80: is this effect triggered when `isBrowser() && window.LC_API` are changed but there is no render?

src/components/containers/show.js
```diff
@@ -30,14 +30,14 @@ export const Mobile = ({ children, ...props }) => (
 export const Eu = ({ children }) => {
     const { is_eu_country } = React.useContext(LocationContext)
 
-    if (is_eu_country) return <>{children}</>
+    if (is_eu_country) return <div>{children}</div>
```
oskar-binary, line 33: The `div` container should not be handled here imo, in case you want to add styling or use a different element.

src/html.js
mustofa-binary: yes, we will remove it from gtm 🙏

src/components/layout/layout.js
mustofa-binary: 👍

src/components/layout/layout.js
mustofa-binary: sort of, its triggered when LC_API changed

oskar-binary: But LC_API changing won't trigger a rerender I think?

src/components/containers/show.js
mustofa-binary: actually, using react fragment makes the children for EU failed to render, @suthesh finds out that its fixed by adding div in show eu component

mustofa-binary: but yes, it should be handled in different pr

src/components/layout/layout.js
mustofa-binary: yes, but `setLiveChatInteractive` does

oskar-binary: but that won't be fired unless the hook is triggered, what about using https://developers.livechat.com/docs/extending-ui/extending-chat-widget/javascript-api/#on-ready after the render []?

ashkanx:
@mustofa-binary make sure it doesn't load on test links. it should load only on ***.deriv.com**

src/components/layout/layout.js
```diff
@@ -121,14 +142,26 @@ const Layout = ({ children, type, interim_type, padding_top, no_login_signup })
...
+            {is_livechat_interactive && (
+                <LiveChat
+                    id="gtm-deriv-livechat"
+                    onClick={() => {
+                        LC_API.open_chat_window()
+                    }}
+                    onMouseEnter={() => setLivechatHover(true)}
+                    onMouseLeave={() => setLivechatHover(false)}
+                >
+                    {is_livechat_hover ? <LiveChatHover /> : <LiveChatIC />}
```
oskar-binary, line 161: can this be handled with css? 🤔 would be faster and more interactive.

mustofa-binary: i thought of that at first, but its a different svg structure

oskar-binary: wouldn't a parent element + display-none on hover for `LiveChatIC` (and reverse) for `LiveChatHover` work?

oskar-binary: actually nvm it's fast enough 👍

PR #971, opened 2020-07-23
oskar/set_cookies

oskar-binary:
Use old cookie for utm_data if it exists and there are no utm params in the url

### DIFF
--- a/src/common/traffic-source.js
+++ b/src/common/traffic-source.js
@@ -64,7 +64,7 @@ const TrafficSource = (() => {
         const makeSetCookie = (object) => (key) =>
             object[key] && cookie.set(key, object[key], { sameSite: 'none', secure: true })
 
-        if (params.utm_source || params.utm_medium || params.utm_campaign) {
+        if (params.utm_source || params.utm_medium || params.utm_campaign || params.t) {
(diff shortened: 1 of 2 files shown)


src/common/traffic-source.js
```diff
@@ -53,14 +61,16 @@ const TrafficSource = (() => {
...
-                    cookie.set(key, params[key], { sameSite: 'none', secure: true })
-                }
-            })
+        const makeSetCookie = (object) => (key) =>
+            object[key] && cookie.set(key, object[key], { sameSite: 'none', secure: true })
+
+        if (params.utm_source || params.utm_medium || params.utm_campaign) {
+            const setParamCookie = makeSetCookie(params)
+            param_keys.forEach(setParamCookie)
+        } else if (existing_utm_cookie) {
```
mustofa-binary, line 70: if it only has affiliate token without utms, then it should have been no utm right?, can we check both if there is affiliate token and utm in the param, if not then we can use the old one

oskar-binary: Do you mean: if the url only has affiliate token -- but no utm params -- we should not use the old cookie?

mustofa-binary: thats right

PR #973, opened 2020-07-24
oskar/gtm_for_test_domain

oskar-binary:
# Description

We'll use a separate GTM instance `deriv-com.binary.sx` --> to use it as a playground for GTM testing.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -31,7 +31,7 @@ export default function HTML(props) {
                         src="https://www.googletagmanager.com/ns.html?id=GTM-NF7884S"
                         height="0"
                         width="0"
-                        style="display:none;visibility:hidden"
+                        style={{ display: 'none', visibility: 'hidden' }}
                     ></iframe>
                 </noscript>
                 <noscript>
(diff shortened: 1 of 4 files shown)


src/html.js
```diff
@@ -11,9 +11,37 @@ export default function HTML(props) {
                     name="viewport"
                     content="width=device-width, initial-scale=1, shrink-to-fit=no"
                 />
+                <script
```
mustofa-binary, line 14: shouldnt we have <script async="" src="//www.googletagmanager.com/gtm.js?id=GTM-XXX"></script> ?

oskar-binary: Yes, thanks!

src/html.js
```diff
@@ -11,9 +12,37 @@ export default function HTML(props) {
...
+                        __html: `
+                    if (window.location.hostname === 'deriv-com.binary.sx') gtm_id = 'GTM-TNX2ZKH'
+                    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+                    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+                    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+                    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+                    })(window,document,'script','dataLayer', gtm_id)`,
+                    }}
+                />
+                <script async src={'https://www.googletagmanager.com/gtm.js?id=' + gtm_id}></script>
```
mustofa-binary, line 26: i m not sure whether it will get the hoisting correctly, but shouldnt this be above the create element script

PR #979, opened 2020-07-27
oskar/response_headers

oskar-binary:
Update Respone header cache-control handling as per https://www.gatsbyjs.org/docs/caching/

### DIFF
--- a/deriv.com.conf
+++ b/deriv.com.conf
@@ -32,7 +32,7 @@ server {
     listen 80;
     server_name  _;
 
-    location ~* \.(html)$ {
+    location ~* \.html$ {
         add_header Cache-Control "public, no-cache, must-revalidate";
     }
 
--- a/deriv.com.conf
+++ b/deriv.com.conf
@@ -32,6 +32,26 @@ server {
     listen 80;
     server_name  _;
 
+    location ~* \.(?:html)$ {
+        add_header Cache-Control "public, max-age=0, must-revalidate";
+    }
+
+    location /page-data {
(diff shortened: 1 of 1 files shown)


deriv.com.conf
```diff
@@ -32,7 +32,26 @@ server {
...
-    add_header Cache-Control "public, max-age=7200, s-maxage=600";
+    location ~* \.html$ {
+        add_header Cache-Control "public, no-cache, must-revalidate";
+    }
+
+    location /page-data {
+        add_header Cache-Control "public, no-cache, must-revalidate";
+    }
+
+    location = /sw.js {
```
mustofa-binary, line 43: is sw.js service worker related?

oskar-binary: yes we never want to cache the service worker - as we it should be revalidated upon each load to check if a new version of the site is available. (once we add the service worker that is :) )

deriv.com.conf
```diff
@@ -32,7 +32,26 @@ server {
...
+        add_header Cache-Control "public, no-cache, must-revalidate";
+    }
+
+    location /static {
+        add_header Cache-Control "public, max-age=31536000, immutable";
```
mustofa-binary, line 48: any reason why keep cache for 1 year?, isnt it too long, why dont put it a day or a week then it will be revalidated?

oskar-binary: These files get a unique hash from webpack (from content) so they are safe to cache forever: check https://www.gatsbyjs.org/docs/caching/

mustofa-binary: nice!, 👍

PR #1014, opened 2020-08-07
amam/improve livechat script performance

mustofa-binary:
# Description

Moving livechat script to onClient entry

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release


gatsby-browser.js
```diff
@@ -44,6 +44,14 @@ export const onInitialClientRender = () => {
 export const onClientEntry = () => {
     LocalStore.set('window_loaded', '')
     window.onload = () => LocalStore.set('window_loaded', 'true')
```
oskar-binary, line 46: Could you add inside the window.onload event?

```javascript
window.onload = () => {
     LocalStore.set('window_loaded', 'true')
     ... add livechat script
}
```

src/html.js
```diff
@@ -17,16 +17,6 @@ export default function HTML(props) {
                 {props.preBodyComponents}
                 <div key={`body`} id="___gatsby" dangerouslySetInnerHTML={{ __html: props.body }} />
                 {props.postBodyComponents}
-                <script
```
oskar-binary, line 20: We don't need src/html.js anymore I think?

mustofa-binary: ya, we dont need it anymore for now, lets remove it

PR #1035, opened 2020-08-17
Sara / Added check domain function to avoid copy 

sara-fs:
# Description

Changes:

-   fixes issue with item
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Refactor code
-   [ ] Breaking change
-   [ ] Dependency update
-   [ ] Release

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -16,11 +16,11 @@ const checkDomain = () => {
         ),
     )
 }
+
 export const onInitialClientRender = () => {
     // Enable translation
     // Check if not production and match ach or ach/
     if (is_browser) {
-        checkDomain()
         const match_ach = window.location.pathname.match(/^(\/ach\/)|\/ach$/)
 
         if (match_ach) {
@@ -79,6 +79,7 @@ export const onPreRouteUpdate = () => {
 
 export const onRouteUpdate = () => {
     NProgress.done()
+    checkDomain()
 }
 
 export const wrapPageElement = WrapPagesWithLocaleContext
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -9,7 +9,7 @@ import 'typeface-ibm-plex-sans'
 
 const is_browser = typeof window !== 'undefined'
 
-function checkDomain() {
+const checkDomain = () => {
     return eval(
         decodeURIComponent(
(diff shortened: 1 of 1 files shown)


gatsby-browser.js
```diff
@@ -7,13 +7,20 @@ import 'typeface-ibm-plex-sans'
...
+    )
+}
 export const onInitialClientRender = () => {
     // Enable translation
     // Check if not production and match ach or ach/
 
     if (is_browser) {
         const match_ach = window.location.pathname.match(/^(\/ach\/)|\/ach$/)
-
+        checkDomain()
```
mustofa-binary, line 23: shouldnt it be 
```
if (!checkDomain()) {
  alert('xxx')
}
```

gatsby-browser.js
```diff
@@ -7,13 +7,20 @@ import 'typeface-ibm-plex-sans'
 
 const is_browser = typeof window !== 'undefined'
 
+function checkDomain() {
```
mustofa-binary, line 10:
```suggestion
const checkDomain = () => {
```

gatsby-browser.js
```diff
@@ -7,13 +7,20 @@ import 'typeface-ibm-plex-sans'
...
+    return eval(
+        decodeURIComponent(
+            'var%20curhost%20%3D%20window.location.hostname%3B%20var%20t8hvj%20%3D%20%2F%5Cb%28deriv%7Cbinary%7Cbinaryqa%5B0-9%5D%7B2%7D%29%5C.%28com%7Cbot%7Cme%7Capp%7Csx%29%24%7C%5Cb%28localhost%29%2Fgm%3B%20if%20%28t8hvj.test%28curhost%29%20%3D%3D%20false%29%7Balert%28%22Not%20our%20domain%22%29%7D',
```
mustofa-binary, line 13: is this from codepen?

gatsby-browser.js
sara-fs: `alert` has been encoded inside the function. Nothing is visible.

sara-fs: This is for increasing the security.

gatsby-browser.js
sara-fs: Aditya suggested `jsfuck.com`, but the encoded result was more than 50000 chaeacters and very long for a simple encoding, so I used this site: `https://www.urlencoder.org/`; Please, suggest if you have any better option. Thanks.

mustofa-binary (approved):

mustofa-binary (approved):

PR #1044, opened 2020-08-23
Arash/add new payments

arash-fs:
# Description

Summary

Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/payment-methods/_expanded-list.js
```diff
@@ -91,12 +91,15 @@ const Withdrawal = styled(Td)`
...
 `
 
-const ExpandList = ({ data }) => {
+const ExpandList = ({ data, config, is_crypto }) => {
     const [is_expanded, setIsExpanded] = React.useState(false)
 
     const toggleExpand = () => {
         setIsExpanded(!is_expanded)
     }
+    function getCryptoConfig(name) {
```
mustofa-binary, line 100: can we standardize this to use arrow function

src/pages/payment-methods/_expanded-list.js
```diff
@@ -91,12 +91,15 @@ const Withdrawal = styled(Td)`
...
+        return config == undefined ? null : config[name].minimum_withdrawal
```
mustofa-binary, line 101:
```suggestion
        return typeof config !== undefined && config[name].minimum_withdrawal
```

src/pages/payment-methods/_payment-data.js
```diff
@@ -417,55 +447,163 @@ const payment_data = [
...
+                min_max_withdrawal: undefined,
+                deposit_time: <Localize translate_text="Funds available as soon as confirmed" />,
+                withdrawal_time: <Localize translate_text="Subject to internal checks" />,
+                description: (
+                    <Localize translate_text="Deposit and withdraw in USDK, a cryptocurrency that’s pegged to USD." />
+                ),
+                name: 'USDK',
+            },
+            {
+                method: <TetherIc />,
```
mustofa-binary, line 585: is the icon for usdt and eusdt is thether?

src/pages/payment-methods/index.js
```diff
@@ -178,15 +182,28 @@ const PaymentMethods = () => {
...
-                                                {pd.data.map((data, indx) => (
-                                                    <ExpandList
-                                                        key={indx}
-                                                        data={data}
-                                                        is_crypto={pd.is_crypto}
-                                                    />
-                                                ))}
-                                            </Tbody>
+                                            <TbodyContainer>
+                                                {(config) => (
```
mustofa-binary, line 186: any reason why is this a function?, and when do we call it, if the intention is the check whether config is available, we can do 
```
{config && <Tbody></Tbody>}
```

mustofa-binary (changes requested):

src/pages/payment-methods/index.js
arash-fs: because we want to use `useContext` so we can access it on the child component of the layout.
since `PaymentMethods` component is actually layout we cannot access to `LocationContext` inside that so, I decided to create a Container then use the `useContext` there.

src/pages/payment-methods/_payment-data.js
arash-fs: it's based on the design

src/pages/payment-methods/_expanded-list.js
arash-fs: done :+1:

src/pages/payment-methods/_expanded-list.js
arash-fs: done :+1:

src/pages/payment-methods/_expanded-list.js
mustofa-binary: i see, how about the condition, is it trying to check for config type or the config itself?

arash-fs: config itself.
basically we are initiating the `config` with the `undefined` and we are checking that.

mustofa-binary: ah, no wonder, okay 👍

mustofa-binary (approved):

mustofa-binary (approved):

PR #1180, opened 2020-11-02
abdalla/fix alignment error

AbdallaMourad:
# Description

Summary

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/_dhero.js
+++ b/src/components/custom/_dhero.js
@@ -137,7 +137,6 @@ const GoToLiveDemo = styled(LinkButton)`
     @media ${device.mobileL} {
         max-width: 100%;
         padding: 10px 11px;
-        height: 42px;
         white-space: nowrap;
         margin-left: 0;
     }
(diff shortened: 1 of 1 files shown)


src/components/custom/_dhero.js
```diff
@@ -128,17 +128,18 @@ const GoToLiveDemo = styled(LinkButton)`
...
     width: auto;
 
     @media ${device.laptopM} {
         font-size: 14px;
-        margin-left: 0;
     }
     @media ${device.mobileL} {
         max-width: 100%;
         padding: 10px 11px;
+        height: 42px;
```
mustofa-binary, line 140: can we use padding only for both buttons, instead of adding fix height?

AbdallaMourad: I saw there was a fixed height for the other button in case of Mobile. If I change it to padding I might need to change it in the other button as well. Will try to look into that ✌🏻🙏🏻

mustofa-binary (approved):

mustofa-binary (approved):

PR #1207, opened 2020-11-18
abdalla/UNIFY: tabs effect

AbdallaMourad:
## Changes Affect the following pages:

- [x] about
- [x] markets
- [x] terms-and condition
- [x] help-centre

### DIFF
--- a/src/components/hooks/use-tab-state.js
+++ b/src/components/hooks/use-tab-state.js
@@ -1,41 +1,32 @@
 import { useState, useEffect } from 'react'
-import { isBrowser, scrollTop } from 'common/utility'
-
-const getHashValue = hash => hash ? hash.slice(1) : null
-const checkTabList = (tab_list, tab) => tab_list.includes(tab)
+import { checkElemInArray, getLocationHash, isBrowser, routeBack, scrollTop, setLocationHash,  } from 'common/utility'
 
 export const useTabState = (tab_list) => {
     const [active_tab, setActiveTab] = useState(tab_list[0])
 
     useEffect(() => {
-        const hash_val = getHashValue(location.hash)
-        
-        if (!location.hash || !checkTabList(tab_list, hash_val)) {
-            location.hash = `#${active_tab}`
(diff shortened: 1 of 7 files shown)


kevinw-binary (dismissed):

src/components/hooks/use-tab-state.js
```diff
@@ -0,0 +1,32 @@
+import { useState, useEffect } from 'react'
+import { checkElemInArray, getLocationHash, isBrowser, routeBack, scrollTop, setLocationHash,  } from 'common/utility'
+
+export const useTabState = (tab_list) => {
+    const [active_tab, setActiveTab] = useState(getLocationHash() ? getLocationHash() : tab_list[0])
```
mustofa-binary, line 5: we only rely on hash when its mounted, i guess we can add getlocation hash inside didmount useEffect. and then listen to active_tab to set location hash scrolltop, or routeback

src/pages/about/index.js
```diff
@@ -157,27 +156,20 @@ const StyledText = styled(Text)`
...
-    const [active_tab, setTab] = useTabState('story')
-    const is_story = active_tab === 'story'
-    const is_leadership = active_tab === 'leadership'
+    const [active_tab, setActiveTab] = useTabState(['story', 'leadership'])
+    const [is_story, setStory] = useState(false)
+    const [is_leadership, setLeadership] = useState(false)
+
+    useEffect(() => {
+        setMobile(isBrowser() && getWindowWidth() <= size.tablet)
+    }, [getWindowWidth()])
```
mustofa-binary, line 167: is this working as expected?

AbdallaMourad: Yes it is

src/components/hooks/use-tab-state.js
AbdallaMourad: This sets the location if the user intendedly enter the URL with the hash provided.

abdalla-binary (approved):

mustofa-binary (approved):

PR #1221, opened 2020-11-23
abdalla/Fix: hero padding issue

AbdallaMourad:
# Description

Summary

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/layout.js
+++ b/src/components/layout/layout.js
@@ -16,7 +16,7 @@ import { DerivStore } from 'store'
 const LiveChat = Loadable(() => import('./livechat'))
 
 const Main = styled.main`
-    padding-top: ${(props) => props.padding_top || '7rem'};
+    top: ${(props) => props.top || '7rem'};
     background: var(--color-white);
     height: 100%;
     position: relative;
@@ -27,7 +27,7 @@ const has_dataLayer = isBrowser() && window.dataLayer
 const TRACKING_STATUS_KEY = 'tracking_status'
 const tracking_status_cookie = new CookieStorage(TRACKING_STATUS_KEY)
 
-const Layout = ({ children, type, interim_type, padding_top, no_login_signup }) => {
+const Layout = ({ children, type, interim_type, top, no_login_signup }) => {
     const { is_eu_country } = React.useContext(DerivStore)
     const [has_mounted, setMounted] = React.useState(false)
     const [show_cookie_banner, setShowCookieBanner] = React.useState(false)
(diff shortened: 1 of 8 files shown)


src/components/layout/layout.js
```diff
@@ -16,7 +16,7 @@ import { DerivStore } from 'store'
 const LiveChat = Loadable(() => import('./livechat'))
 
 const Main = styled.main`
-    padding-top: ${(props) => props.padding_top || '7rem'};
+    top: ${(props) => props.top || '7rem'};
```
mustofa-binary, line 19: can you help to explain why?

AbdallaMourad: Padding creates white space which will be covered by nav bar. it will show only when you scroll down on ios devices. Since Hero section needs have a space between the nav and content. `top` property appears to be a correct solution.

src/pages/careers/index.js
```diff
@@ -10,7 +10,7 @@ import Layout from 'components/layout/layout'
 import { localize, WithIntl } from 'components/localization'
 
 const Careers = () => (
-    <Layout type="careers" padding_top="7rem">
+    <Layout type="careers" margin_top="7rem">
```
sean-binary, line 13: The issue seems to have been fixed 👍 . Just wanted some clarification, when do you use `margin top` and when do you use `top`?

sean-binary (approved):

src/pages/careers/index.js
abdalla-binary: `margin-top` is when you want to make a space between the selected container and the one above it. It won't affect the other containers since all of them will be shifted with the same amount of pixels. `top` is to shift the container by amount of pixels but it will affect the other containers.

abdalla-binary: to be more clearer see the [deployment of `top`](https://deriv-com-5b96mb25u.binary.sx/) and [deployment of `margin-top`](https://deriv-com-h021vpwou.binary.sx/) notice that the logo in footer is covered by the element above it. where it doesn't occur in the latter.

src/components/layout/layout.js
abdalla-binary: > `Padding` creates white space which will be covered by nav bar. it will show only when you scroll down on ios devices. Since Hero section needs have a space between the nav and content. `top` property appears to be a correct solution.

`fixed in the latter commit`

mustofa-binary (dismissed):

mustofa-binary (approved):

PR #1311, opened 2020-12-28
Yashim/fix underline navbar

yashim-deriv:
# Description
Summary
- Main navigation bar now shows current parent page by displaying a red underline under the link

Changes:
- Moved all card links to an array
- If current page matches a link in the array, return a string containing current navigation dropdown name
- Check links for the current page string. If match, sets the `is_current_page` to true which will activate the red underline.

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/other-platforms.js
+++ b/src/components/custom/other-platforms.js
@@ -13,7 +13,7 @@ import {
 } from 'components/elements'
 import { localize, LocalizedLink, Localize } from 'components/localization'
 import device from 'themes/device'
-import { community_url, isBrowser } from 'common/utility'
+import { community_url, getLocationHash } from 'common/utility'
 // icons
 import DTrader from 'images/svg/dtrader-icon.svg'
 import DMT5 from 'images/svg/dmt5-icon.svg'
@@ -336,7 +336,7 @@ export const NavMarket = ({ onClick, setCurrentPage }) => {
     useEffect(() => {
         if (Object.values(links).includes(window.location.pathname + window.location.hash))
             setCurrentPage('markets')
-    }, [isBrowser() ? window.location.hash : null])
+    }, [getLocationHash()])
 
     return (
         <Flex direction="column" wrap="wrap" jc="flex-start">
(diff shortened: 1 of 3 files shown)


src/components/custom/other-platforms.js
```diff
@@ -224,226 +224,289 @@ OtherPlatform.propTypes = {
...
+        dmt5: '/dmt5',
+        smarttrader: 'trading',
+        margin: '/trade-types/margin',
+        options: '/trade-types/options',
+        multipliers: '/trade-types/multiplier',
+    }
+
+    useEffect(() => {
+        if (Object.values(links).includes(window.location.pathname)) setCurrentPage('trade')
+    }, [])
```
sean-binary, line 240: I think you'll need to add a dependency here to check when the hash changes. Cause right now, if I were to key in `/about` it will automatically reroute to `/about#story` without updating the state, hence not showing the active bar. Same goes for all the rest.

yashim-deriv: Good catch~ I'll check this out. Thanks!

src/components/custom/other-platforms.js
```diff
@@ -336,7 +336,7 @@ export const NavMarket = ({ onClick, setCurrentPage }) => {
     useEffect(() => {
         if (Object.values(links).includes(window.location.pathname + window.location.hash))
             setCurrentPage('markets')
-    }, [window?.location.hash])
+    }, [isBrowser() ? window.location.hash : null])
```
AbdallaMourad, line 339: There is a function in `utility` that has the same implementation called `getLocationHash` it's better to use it here

src/components/custom/other-platforms.js
```diff
@@ -224,226 +224,289 @@ OtherPlatform.propTypes = {
...
+        choose: '/why-choose-us/',
+        partners: '/partners/',
+        contact: '/contact-us/',
+        career: '/careers/',
+    }
+
+    useEffect(() => {
+        if (Object.values(links).includes(window.location.pathname + window.location.hash))
+            setCurrentPage('company')
+    }, [isBrowser() ? window.location.hash : null])
```
AbdallaMourad, line 404: Same here

src/components/custom/other-platforms.js
yashim-deriv: I just noticed this. Will change, thanks 😄

PR #1313, opened 2020-12-28
priojeet/fix:redirect-is-not-properly-set-for-kid

priojeet-deriv:
# Description

Summary

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/localization/localized-link.js
```diff
@@ -157,10 +157,10 @@ export const LocalizedLink = React.forwardRef(({ to, ...props }, ref) => {
             </ExternalLink>
         )
 
-    // internal links should end with / e.g. /about/
-    let internal_to = path_to.charAt(to.length - 1) === '/' ? path_to : path_to
+    let internal_to = path_to
```
sean-binary, line 160:
```suggestion
```

src/components/localization/localized-link.js
```diff
@@ -157,10 +157,10 @@ export const LocalizedLink = React.forwardRef(({ to, ...props }, ref) => {
...
 
-    if (props.has_no_end_slash) {
+    // remove trailing character only if it is a forward slash
+    if (props.has_no_end_slash && internal_to.charAt(internal_to.length - 1) == '/') {
         internal_to = internal_to.substring(0, internal_to.length - 1)
     }
```
sean-binary, line 165:
```suggestion
    } else {
      let internal_to = path_to
    }
```

sean-binary: Just a suggestion, but would it be more concise to declare it like this?

priojeet-deriv: won't we have to define `internal_to` above as let is block scoped?

sean-binary: Oh good point, we could instead use a ternary declaration then. Otherwise, it's fine just as it is.

priojeet-deriv: tried ternary, looks a little congested. more expressive this way.

sean-binary (approved):

mustofa-binary (approved):

PR #1329, opened 2020-12-31
Mahdi Aryayi / Replace normal IBM plex font with variable font

mahdiaryayi-fs:
# Description

We are using the "IBM Plex Sans" font as our main font. the issue is we need to fetch a new font file for every new font-weight we use. but thanks to [Variable Fonts](https://en.wikipedia.org/wiki/Variable_fonts), now we can fetch only one file that supports all font weights (from 100 to 700)

Changes:

-   The new IBM Plex variable font files and related CSS are added to the repository
- the former `typeface-ibm-plex-sans` package is removed from Deriv-com.
- some auto vulnerability fixes applied with `npm audit fix`

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [x] Improve performance
-   [x] Style only
-   [x] Dependency update
-   [ ] Documentation update
-   [ ] Release


static/css/ibm-plex-sans-var.css
```diff
@@ -0,0 +1,17 @@
+@font-face {
+    font-family: 'IBM Plex Sans';
+    font-style: normal;
+    font-display: swap;
+	src:url('../fonts/IBMPlexSansVar-Roman.woff2') format('woff2 supports variations'),
+        url('../fonts/IBMPlexSansVar-Roman.woff2') format('woff2-variations');
+	font-weight: 100 700;
```
AbdallaMourad, line 7: We are also using font-weight `200-300-500-600` do we need to install them and import/define them here?

mahdiaryayi-fs: > We are also using font-weight `200-300-500-600` do we need to install them and import/define them here?

No, `font-weight: 100 700;` is a range. it means all weights from 100 to 700. (they don't even need to be multiple of 100, for example, we can have: `font-weight: 456;`)

AbdallaMourad: I see. Thank you for the clarification 😄🙏🏻

AbdallaMourad (approved):

kevinw-binary (commented):
Do we need to remove the package-lock.json content for this PR?

mahdiaryayi-fs:
> Do we need to remove the package-lock.json content for this PR?

That's not directly related to this task. I just did a `npm audit fix` to fix dependency vulnerabilities. I mentioned that in the PR description

salarhfz-fs (approved):

suthesan-deriv (approved):

PR #1336, opened 2021-01-05
Priojeet options and cfd warnings

mustofa-binary:
# Description

Summary

Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/layout.js
+++ b/src/components/layout/layout.js
@@ -69,6 +69,7 @@ const Layout = ({
     no_login_signup,
     no_live_chat,
     nav_type,
+    CompotentAbove,
 }) => {
     const { is_eu_country } = React.useContext(DerivStore)
     const [has_mounted, setMounted] = React.useState(false)
@@ -154,7 +155,7 @@ const Layout = ({
             FooterNav = <Copyright />
             break
         default:
-            Navigation = <Nav />
+            Navigation = <Nav CompotentAbove={CompotentAbove} />
             FooterNav = <Footer />
             break
     }
@@ -203,6 +204,7 @@ const Layout = ({
 
 Layout.propTypes = {
     children: PropTypes.node.isRequired,
+    CompotentAbove: PropTypes.elementType,
     interim_type: PropTypes.string,
     margin_top: PropTypes.number,
     nav_type: PropTypes.string,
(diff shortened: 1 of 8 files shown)


src/components/layout/layout.js
```diff
@@ -24,32 +25,53 @@ const has_dataLayer = isBrowser() && window.dataLayer
 const TRACKING_STATUS_KEY = 'tracking_status'
 const tracking_status_cookie = new CookieStorage(TRACKING_STATUS_KEY)
 
-const cfdWarningHeight = 8
+const cfd_warning_height_desktop = 8
+const cfd_warning_height_tablet = 12
```
sean-binary, line 29: Since you've declared them in utils, you can just import them instead of declaring it here again.

src/pages/signup/index.js
```diff
@@ -66,7 +66,7 @@ const NewSignup = () => {
         setSubmitState(submitStatus)
     }
     return (
-        <Layout type="static" margin_top={0}>
+        <Layout type="static" margin_top={'0'}>
```
sean-binary, line 69: Should `0` be a number instead of a string?

src/components/layout/layout.js
mustofa-binary: @priojeet-deriv pls take note, will do this in the next PR

src/pages/signup/index.js
mustofa-binary: there were usage where
```
margin_top && margin_top
```

if we use 0, it will return false, where '0' will be the correct props

sean-binary: May need to adjust the PropType in the next PR as well.

sean-binary (approved):

kevinw-binary (approved):

src/components/layout/layout.js
priojeet-deriv: nice catch. will do.

src/pages/signup/index.js
priojeet-deriv: should it be changed to string or accept both string/number?

sean-binary: You can set it to accept both 👍

priojeet-deriv: noted.

PR #1363, opened 2021-01-19
Yashim/fix: getLanguage() returning invalid language codes

yashim-deriv:
Co-authored-by: Yashim Wong <yashimwong@gmail.com>

Summary
- getLanguage() was returning wrong language codes (kebab-case) 

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/common/utility.js
+++ b/src/common/utility.js
@@ -62,7 +62,8 @@ const checkElemInArray = (tab_list, tab) => tab_list.includes(tab)
 
 const getWindowWidth = () => (isBrowser() && window.screen ? window.screen.width : '')
 
-const getLanguage = () => (isBrowser() ? localStorage.getItem('i18n') || navigator.language : null)
+const getLanguage = () =>
+    isBrowser() ? localStorage.getItem('i18n')?.replace('-', '_') || navigator.language : null
 
 const getCrowdin = () =>
     isBrowser() ? localStorage.getItem('jipt_language_code_deriv-com') || navigator.language : null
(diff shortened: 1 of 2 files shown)


AbdallaMourad (approved):

src/common/utility.js
```diff
@@ -62,7 +62,8 @@ const checkElemInArray = (tab_list, tab) => tab_list.includes(tab)
 
 const getWindowWidth = () => (isBrowser() && window.screen ? window.screen.width : '')
 
-const getLanguage = () => (isBrowser() ? localStorage.getItem('i18n') || navigator.language : null)
+const getLanguage = () =>
```
mustofa-binary, line 65: this will break _signup-new , as it needs `-` for redirecting the language, lets not change it in utility, we can use replace function specifically in socket_base, or make another function specifically to replace the case and use it in socket base only

mustofa-binary (changes requested):

src/common/utility.js
yashim-deriv: Yes, I just realised this will break the `/${getLanguage()}/terms-and-conditions/` redirection rule. Will fix right away.

mustofa-binary (approved):

PR #1368, opened 2021-01-20
Priojeet/fix:deriv-careers-page-width-is-greater-than-viewport

priojeet-deriv:
# Description

Summary

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/nav.js
+++ b/src/components/layout/nav.js
@@ -117,8 +117,7 @@ export const Wrapper = styled(Container)`
         font-size: var(--text-size-xxs);
     }
     @media ${device.mobileM} {
-        ${({ offset_px_mobile }) =>
-            offset_px_mobile ? `width: calc(100% - ${offset_px_mobile}px)` : ``};
+        ${({ offset_px_mobile }) => offset_px_mobile && `width: calc(100% - ${offset_px_mobile}px)`};
     }
 `
 
--- a/src/components/layout/nav.js
+++ b/src/components/layout/nav.js
@@ -116,6 +116,9 @@ export const Wrapper = styled(Container)`
     @media ${device.laptop} {
         font-size: var(--text-size-xxs);
     }
+    @media ${device.mobileM} {
+        width: 98%;
+    }
 `
 export const NavLeft = styled.div`
     text-align: left;
--- a/src/components/layout/nav.js
+++ b/src/components/layout/nav.js
@@ -117,9 +117,15 @@ export const Wrapper = styled(Container)`
         font-size: var(--text-size-xxs);
     }
     @media ${device.mobileM} {
-        width: ${(props) => `calc(100% - ${props.offset_px_mobile}px)`};
+        ${({ offset_px_mobile }) =>
+            offset_px_mobile ? `width: calc(100% - ${offset_px_mobile}px)` : ``};
     }
 `
+
+Wrapper.propTypes = {
+    offset_px_mobile: PropTypes.number,
+}
+
 export const NavLeft = styled.div`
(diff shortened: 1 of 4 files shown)


src/components/layout/nav.js
```diff
@@ -116,6 +116,9 @@ export const Wrapper = styled(Container)`
     @media ${device.laptop} {
         font-size: var(--text-size-xxs);
     }
+    @media ${device.mobileM} {
+        width: 98%;
```
sean-binary, line 120: Could this be replaced with a `calc(100%-offset_px_required)` or it is more suitable in this case for it to be in a specific `%`?

priojeet-deriv: updated. thanks :)

mustofa-binary (commented):
@priojeet-deriv there are two commits that is not belong to this PR, can you try to pull off those two commits by rebasing to the latest master?

1d323dc
91d44f6

src/components/layout/nav.js
```diff
@@ -116,7 +116,16 @@ export const Wrapper = styled(Container)`
     @media ${device.laptop} {
         font-size: var(--text-size-xxs);
     }
+    @media ${device.mobileM} {
+        ${({ offset_px_mobile }) =>
+            offset_px_mobile ? `width: calc(100% - ${offset_px_mobile}px)` : ``};
```
sean-binary, line 121:
```suggestion
        ${({ offset_px_mobile }) => offset_px_mobile && `width: calc(100% - ${offset_px_mobile}px)`};
```

sean-binary: IMO, I think we should write it in this manner as a form of better practice because if no prop was passed, we wouldn't want to return `width: ''` as an invalid CSS before falling back to whatever value it was before.

priojeet-deriv: it won't return `width: ''` as the whole expression is wrapped after the condition. But we can do like this to maintain consistency.

src/pages/regulatory/index.js
```diff
@@ -239,6 +239,10 @@ const Regulatory = () => {
                                                 target="_blank"
                                                 to="https://beta.gamblingcommission.gov.uk/public-register/business/detail/39495"
                                                 rel="noopener noreferrer"
+                                                style={{
+                                                    color: `var(--color-black-3)`,
+                                                    textDecoration: `none`,
```
AbdallaMourad, line 244: Nit: Can we use `''` instead of backticks

src/pages/regulatory/index.js
```diff
@@ -282,6 +286,10 @@ const Regulatory = () => {
                                                 target="_blank"
                                                 to="https://beta.gamblingcommission.gov.uk/public-register/business/detail/39172"
                                                 rel="noopener noreferrer"
+                                                style={{
+                                                    color: `var(--color-black-3)`,
+                                                    textDecoration: `none`,
```
AbdallaMourad, line 291: Same here

sean-binary (approved):

AbdallaMourad (approved):

src/pages/regulatory/index.js
priojeet-deriv: done.

yashim-deriv (approved):

PR #1390, opened 2021-01-29
Sean / replaced useStaticQuery to StaticQuery

sean-binary:
Changes:

-   Replaced useStaticQuery to StaticQuery temporarily to solve "useStaticQuery could not be fetched error"

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/dmt5/index.js
+++ b/src/pages/dmt5/index.js
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react'
-import { graphql, useStaticQuery } from 'gatsby'
+import { graphql, StaticQuery } from 'gatsby'
 import {
     WhyTrader,
     StartTrader,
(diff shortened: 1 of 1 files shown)


src/pages/dmt5/index.js
```diff
@@ -88,12 +79,23 @@ const DMT5 = () => {
...
-                title={<Localize translate_text="Get into the DMT5 experience" />}
-                data={data}
+            <StaticQuery
+                query={graphql`
+                    query {
+                        deriv_platform: file(relativePath: { eq: "dmt5-banner.png" }) {
+                            ...fadeIn
+                        }
+                    }
+                `}
```
AbdallaMourad, line 89: We can keep query defined at the top of the file, then we replace the current query with `query={query}` to be more readable

suthesan-deriv (approved):

AbdallaMourad (commented):
We need to check if this error occurs in `dtrader` page as well since both pages share the same logic

mustofa-binary (approved):

PR #1391, opened 2021-02-01
Kevin/feat: handle missing chat_id

kevinw-binary:
# Description

To fix missing **chat_id** unhandled promise on

https://my.trackjs.com/details/f3b68322be99485bb74d3cbbb95c5217
Summary

Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/hooks/use-livechat.js
+++ b/src/components/hooks/use-livechat.js
@@ -126,7 +126,7 @@ export const useLivechat = () => {
                         }
                     } else {
                         if (window.LiveChatWidget.get('chat_data')) {
-                            const chatID = window.LiveChatWidget.get('chat_data').chatId || ''
+                            const chatID = window.LiveChatWidget.get('chat_data').chatId ?? ''
                             customerSDK?.deactivateChat({ chatId: chatID }).catch((error) => {
                                 // eslint-disable-nextline
                                 console.error(error)
--- a/src/components/hooks/use-livechat.js
+++ b/src/components/hooks/use-livechat.js
@@ -126,7 +126,7 @@ export const useLivechat = () => {
                         }
                     } else {
                         if (window.LiveChatWidget.get('chat_data')) {
-                            const chatID = window.LiveChatWidget.get('chat_data').chatId
+                            const chatID = window.LiveChatWidget.get('chat_data').chatId || ''
                             customerSDK?.deactivateChat({ chatId: chatID }).catch((error) => {
                                 // eslint-disable-nextline
                                 console.error(error)
--- a/src/components/hooks/use-livechat.js
+++ b/src/components/hooks/use-livechat.js
@@ -125,7 +125,8 @@ export const useLivechat = () => {
                             )
                         }
                     } else {
-                        if (window.LiveChatWidget.get('chat_data')) {
+                        const chat_data = window.LiveChatWidget.get('chat_data')
+                        if (chat_data) {
                             const chatID = window.LiveChatWidget.get('chat_data').chatId ?? ''
                             customerSDK?.deactivateChat({ chatId: chatID }).catch((error) => {
                                 // eslint-disable-nextline


src/components/hooks/use-livechat.js
```diff
@@ -126,7 +126,7 @@ export const useLivechat = () => {
                         }
                     } else {
                         if (window.LiveChatWidget.get('chat_data')) {
-                            const chatID = window.LiveChatWidget.get('chat_data').chatId
+                            const chatID = window.LiveChatWidget.get('chat_data').chatId ?? ''
```
sean-binary, line 129:
```suggestion
                            const chatID = window.LiveChatWidget.get('chat_data')?.chatId
```

sean-binary: Could we use optional chaining here instead?

suthesan-deriv: I doubt we can actually...cause optional chaining would mean that there is a possibility that nothing is passed. From what i understand there was an error on console because nothing was passed for chatID. Maybe @kevinw-binary can confirm on this?

kevinw-binary: There's a chance that `window.LiveChatWidget.get('chat_data')` isn't  defined yet which will cause the error

or is it better to implement `try and catch` since this code is using API to get the data ?

AbdallaMourad: I guess the current logic is correct with a fallback, since we have to assign a value to  chat_id, so `chat_id: undefined` is not an option.

But we need to make sure that `chat_id: ''` is acceptable @kevinw-binary please confirm 🙏🏻

kevinw-binary: I think it will catch the error on the line below here  

`customerSDK?.deactivateChat({ chatId: chatID }).catch((error) => { `

and the fallback will supposedly fix the error from trackJS ?

AbdallaMourad: I believe that the issue appears because we have [`console.error`](https://github.com/binary-com/deriv-com/pull/1391/files#diff-e5e90c60d24e7dae6576024383c266127f2d7aef9cb1717329e7e557cc98fe9fR132) 

A fallback should prevent going to `catch` block, but I am not very sure if an empty string would do it.

AbdallaMourad: As per documentation [`chatId`](https://developers.livechat.com/docs/extending-chat-widget/customer-sdk/#deactivatechat) needs to be of type `string` therefore it's better to have `''` as a fallback to avoid `chatId: undefind`

AbdallaMourad (dismissed):

src/components/hooks/use-livechat.js
```diff
@@ -126,7 +126,7 @@ export const useLivechat = () => {
                         }
                     } else {
                         if (window.LiveChatWidget.get('chat_data')) {
```
yashim-deriv, line 128:
```suggestion
                        const chat_data = window.LiveChatWidget.get('chat_data')
                        if (chat_data) {
```

src/components/hooks/use-livechat.js
```diff
@@ -126,7 +126,7 @@ export const useLivechat = () => {
...
-                            const chatID = window.LiveChatWidget.get('chat_data').chatId
+                            const chatID = window.LiveChatWidget.get('chat_data').chatId ?? ''
                             customerSDK?.deactivateChat({ chatId: chatID }).catch((error) => {
```
yashim-deriv, line 130:
```suggestion
                            customerSDK?.deactivateChat({ chatId: chat_data.chatId ?? '' }).catch((error) => {
```

yashim-deriv (approved):
Just a small suggestion. The current code is already ok though

src/components/hooks/use-livechat.js
mustofa-binary: either that, or call deactivateChat if only there is valid chatId

mustofa-binary (approved):

PR #1452, opened 2021-02-18
Sean / perf: reduce painting from dropdown animations

sean-binary:
Changes:

-   Reduce painting from dropdown animations

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [x] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/platforms-dropdown.js
+++ b/src/components/custom/platforms-dropdown.js
@@ -37,6 +37,7 @@ const NavDropdown = styled.div`
     box-shadow: 0 16px 20px 0 rgba(0, 0, 0, 0.1);
     border-radius: 0.4rem;
     transition: opacity 0.3s, transform 0.3s;
+    will-change: display, left, opacity;
     animation-name: ${(props) => (props.is_open ? FadeInDown : FadeOutUp)};
     animation-fill-mode: both;
     animation-duration: ${(props) => (props.has_animation ? '0.3s' : '0')};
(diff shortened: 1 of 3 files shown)


suthesan-deriv (approved):

yashim-deriv (commented):
Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

Quoting the docs:

>**Don't apply will-change to elements to perform premature optimization**. If your page is performing well, don't add the will-change property to elements just to wring out a little more speed. will-change is intended to be used as something of a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems. Excessive use of will-change will result in excessive memory use and will cause more complex rendering to occur as the browser attempts to prepare for the possible change. This will lead to worse performance.

The proper usage imho is to attach the will change optimisation on event (aka only during change) as per the docs:

```var el = document.getElementById('element');

// Set will-change when the element is hovered
el.addEventListener('mouseenter', hintBrowser);
el.addEventListener('animationEnd', removeHint);

function hintBrowser() {
  // The optimizable properties that are going to change
  // in the animation's keyframes block
  this.style.willChange = 'transform, opacity';
}

function removeHint() {
  this.style.willChange = 'auto';
}```

In this case we should attach it to the event handler that will trigger the animation and remove it once the event is over.

PR #1615, opened 2021-04-12
suthesh/h1-tag-seo

suthesan-deriv:
Changes:

-   replace h1 with span in header

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/other-platforms.js
+++ b/src/components/custom/other-platforms.js
@@ -237,7 +237,6 @@ export const NavPlatform = ({ onClick, is_ppc, is_ppc_redirect }) => {
             <Flex direction="column" wrap="wrap" jc="flex-start">
                 <StyledText>{localize('Trading platforms')}</StyledText>
                 <NavCard
-                    as="span"
                     aria_label="Dtrader"
                     icon={() => <img src={DTrader} alt="Dtrader" width="32" height="32" />}
                     content={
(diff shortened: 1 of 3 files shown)


yashim-deriv:
Since it seems that `as` has the value of `span` in the majority of `NavCard` props, isn't it better to just set the default value of `as="span"` in that component itself?

src/components/custom/other-platforms.js
```diff
@@ -237,6 +237,7 @@ export const NavPlatform = ({ onClick, is_ppc, is_ppc_redirect }) => {
             <Flex direction="column" wrap="wrap" jc="flex-start">
                 <StyledText>{localize('Trading platforms')}</StyledText>
                 <NavCard
+                    as="span"
```
sean-binary, line 240: I don't think you can use the `as` prop here. You can either add `.attrs({ as: 'span' })` directly onto the global `NavCard` and `CardLink` component, or you could extend those onto this page here and add `.attrs({ as: 'a' })` to it.

suthesan-deriv: Actually the `as` prop is passed to NavCard -> ResponsiveHeader -> Header . Technically it works but I think it's better to change the prop name to `title_as` cause it's kinda funny to suddenly see an `as` prop in a Card component? 😂

suthesan-deriv:
> Since it seems that `as` has the value of `span` in the majority of `NavCard` props, isn't it better to just set the default value of `as="span"` in that component itself?

I thought of that too actually, in fact that's what I did in my first commit 😂 ...then thought what if we in the future decide to use that NavCard and CardLink component else where? Do you think that might happen? 🤔

sean-binary (approved):

PR #1619, opened 2021-04-12
Habib/update multiplier RTP

ShuvoHabib:
Changes:

RTP copy: "Return to player for mulitplier options is in the range of 95.0-99.9% for all indices on an average. Using a different multiplier or duration may affect the RTP."

/trade-types/multiplier
/markets#synthetic

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/trade-types/multiplier/_synthetic-indices.js
```diff
@@ -166,6 +166,13 @@ const SyntheticIndices = () => {
                     />
                 </MarketsWrapper>
                 <AvailablePlatforms dtrader />
+                <Show.Eu>
+                    {synthetic_multiplier.eu_content?.map((text, index) => (
```
yashim-deriv, line 170: Everything looks good~ Only a question, will there be any instance that eu_content might return null or undefined?

ShuvoHabib: It won't be null or undefined unless someone renames the key of remove it. It's the same content used in two different pages.

yashim-deriv: If that's the case, i think it's better if we remove the optional chaining here since the value is static, what do you think? 😄

ShuvoHabib: Sounds good to me. Resolved.

yashim-deriv (approved):
LGTM~

sean-binary (approved):

PR #1646, opened 2021-04-19
fiona/ resize hover arrow

fiona-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [x] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/payment-methods/_expanded-list.js
+++ b/src/pages/payment-methods/_expanded-list.js
@@ -14,8 +14,9 @@ const StyledButton = styled(Button)`
 `
 
 const StyledChevron = styled.img`
-    height: 14px;
-    width: 14px;
+    height: 16px;
+    width: 16px;
+    margin-left: 40px;
     transform: ${(props) => (props.expanded ? 'inherit' : 'rotate(-180deg)')};
     transition: transform 0.25s ease-out;
 `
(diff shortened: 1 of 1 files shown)


arome-binary (dismissed):
Chevron sign has been aligned in the center of the hover cell

src/pages/payment-methods/_expanded-list.js
```diff
@@ -16,6 +16,7 @@ const StyledButton = styled(Button)`
 const StyledChevron = styled.img`
     height: 16px;
     width: 16px;
+    margin-left: 40px;
```
sean-binary, line 19: Nice job, however I think that the chevron will no longer be centered on smaller screen widths such as mobile devices as the column widths are slightly different there. Hence, a predetermined 40px margin might not be the best option to cover all cases. A suggestion would be to use `display: flex` for the parent component and to have its `justify-content` and `align-items` set to center.

fiona-deriv: Thanks, I'll work on that

sean-binary (approved):

PR #1650, opened 2021-04-21
yashim/fix/refeactor/feat: add dropdown search to margin calculator

yashim-deriv:
Changes:
- Removed deprecated `KeyboardEvent.keyCode` and replaced with updated `KeyboardEvent.key`. Ref: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
- Refactored repeated functions to `utilities`
- Fixed not working `Dropdown` behaviour (ie, escape, button up and down, etc)
- Refactored `Dropdown` behaviour into a hook to be reusable
- Added `DropdownSearch` component extended from the `Dropdown` component

## Type of change

-   [x] Bug fix
-   [x] New feature
-   [x] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/elements/dropdown.js
```diff
@@ -198,10 +203,9 @@ const Arrow = styled(Chevron)`
     }
 `
 
-const StyledLabel = styled.label`
-    /* prettier-ignore */
-    color: var(--color-${(props) => props.labelColor || 'grey'});
-    background: var(--color-${(props) => props.labelColor || 'white'});
+export const StyledLabel = styled.label`
+    color: grey;
+    background: white;
```
sean-binary, line 208: Let's use our global variables here 🙏

yashim-deriv: Done 😄

src/components/hooks/dropdown-hooks.js
```diff
@@ -0,0 +1,67 @@
...
+
+export const useDropdownHooks = (onChange) => {
+    const [is_open, setOpen] = useState(false)
+    const dropdown_ref = useRef(null)
+    const nodes = new Map()
+
+    useOutsideClick(dropdown_ref, () => setOpen(false))
+
+    const toggleListVisibility = (e) => {
+        const open_dropdown = e.key === 'Space' || e.key === 'Enter' || e.key === 'Tab'
```
mustofa-binary, line 12: there are some machines which the key is a number, can we bring keycodes utils back?

src/common/utility.js
```diff
@@ -67,6 +67,9 @@ export const routeBack = () => {
 }
 export const checkElemInArray = (tab_list, tab) => tab_list.includes(tab)
 
+export const getCommaSeparatedNumber = (input) =>
+    input.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',')
```
mustofa-binary, line 71: please add a comment here 🙏

mustofa-binary (changes requested):

src/common/utility.js
yashim-deriv: Done 😄

src/components/hooks/dropdown-hooks.js
yashim-deriv: there was a deprecation notice for the `keyCode` property: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode which why i used `key` instead

yashim-deriv: Tested with different device. Should work as before.

Compatibility is good but currently only IE 11 is not supported. If you want me do make it compatible with IE 11 i could use a mapper to use the newer `key` property with a code mapper as per this SO post (thread no. 3):

https://stackoverflow.com/questions/35394937/keyboardevent-keycode-deprecated-what-does-this-mean-in-practice

mustofa-binary (approved):

sean-binary (approved):

PR #1694, opened 2021-05-06
suthesh/remove-careers-sitemap

suthesan-deriv:
Changes:

-   removed localized career page links from sitemap

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -53,7 +53,6 @@ module.exports = {
                     '/**/landing',
                     '/**/landing/**',
                     '/**/careers',
-                    '/**/careers/**',
                 ],
                 serialize: ({ site, allSitePage }) =>
                     allSitePage.edges.map((edge) => {
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -52,10 +52,10 @@ module.exports = {
                     '/landing/**',
                     '/**/landing',
                     '/**/landing/**',
-                    '/**/careers/**',
                 ],
                 serialize: ({ site, allSitePage }) =>
                     allSitePage.edges.map((edge) => {
+                        const ignore_localized_regex = /career/;
                         const path = edge.node.path
                         let priority = 0.7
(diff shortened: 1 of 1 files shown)


sean-binary (dismissed):

gatsby-config.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
                 ],
                 serialize: ({ site, allSitePage }) =>
                     allSitePage.edges.map((edge) => {
+                        const ignore_localized_regex = /careers/;
```
sean-binary, line 58:
```suggestion
                        const careers_regex = /careers/;
```

sean-binary: Maybe we can call this by its name perhaps? And then using it in `ignore_localized`

gatsby-config.js
```diff
@@ -82,14 +83,15 @@ module.exports = {
 
                         languages.push('x-default')
                         languages.splice(languages.indexOf('ach'), 1)
-                        const links = languages.map((locale) => {
+                        const ignore_localized = current_page.match(ignore_localized_regex);
+                        let links;
+
+                        links = languages.map((locale) => {
```
sean-binary, line 89:
```suggestion
                        const links = languages.map((locale) => {
```

sean-binary: Can we declare it here instead?

gatsby-config.js
suthesan-deriv: Was planning on using this to probably add more pages into this list in future

gatsby-config.js
suthesan-deriv: Yup we totally should 😂

PR #1719, opened 2021-05-16
Amam/websocket page

mustofa-binary:
Changes:

- new endpoint page
- refactor hooks
- modify deriv store

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [x] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/hooks/use-website-status.js
```diff
@@ -0,0 +1,39 @@
...
+    })
+    const [is_loading, setLoading] = useState(true)
+
+    useEffect(() => {
+        setLoading(true)
+        if (!website_status) {
+            const binary_socket = BinarySocketBase.init()
+
+            binary_socket.onopen = () => {
+                binary_socket.send(JSON.stringify({ website_status: 1 }))
```
ShuvoHabib, line 20: Let's put '1' in a variable, that tells us, what happens when the status is '1'

src/pages/endpoint/index.js
```diff
@@ -0,0 +1,200 @@
...
+    const endpointSubmission = (values, actions) => {
+        actions.setSubmitting(true)
+        setServerUrl(values.server_url)
+        setAppId(values.app_id)
+
+        // handle website status changes
+        const new_website_status = { ...website_status, clients_country: values.clients_country }
+        setWebsiteStatus(new_website_status)
+        actions.setSubmitting(false)
+        handleStatus(actions.setStatus, 'Config has been updated')
```
ShuvoHabib, line 86: Should this text be localized?

src/pages/endpoint/index.js
```diff
@@ -0,0 +1,200 @@
...
+        // TODO: if there is a change requires reload in the future
+        // window.location.reload()
+    }
+
+    return (
+        <Layout type="static" margin_top={'0'}>
+            <SEO title="Endpoint" description="Change deriv API endpoint." no_index />
+            <StyledContainer justify="center" align="center" direction="column">
+                <Header as="h2" type="page-title" align="center" mt="80px">
+                    Change API endpoint
```
ShuvoHabib, line 96: Should this be localized?

src/pages/endpoint/index.js
```diff
@@ -0,0 +1,200 @@
...
+                </Header>
+                <Header
+                    as="h4"
+                    type="sub-section-title"
+                    align="center"
+                    weight="500"
+                    mt="0.5rem"
+                    mb="3.8rem"
+                >
+                    Update configuration for API endpoint or other settings
```
ShuvoHabib, line 106: Should this be localized?

src/pages/endpoint/index.js
```diff
@@ -0,0 +1,200 @@
...
+    return errors
+}
+
+const Endpoint = () => {
+    const [server_url, setServerUrl] = useLocalStorageState(default_server_url, 'config.server_url')
+    const [app_id, setAppId] = useLocalStorageState(getAppId(), 'config.app_id')
+    const { website_status, setWebsiteStatus, website_status_loading } = React.useContext(
+        DerivStore,
+    )
+    const TIMOUT_DELAY = 1500
```
ShuvoHabib, line 61: Can we make it, "TIMEOUT_DELAY"

src/pages/endpoint/index.js
```diff
@@ -0,0 +1,200 @@
...
+    if (app_id_error) {
+        errors.app_id = app_id_error
+    }
+
+    return errors
+}
+
+const Endpoint = () => {
+    const [server_url, setServerUrl] = useLocalStorageState(default_server_url, 'config.server_url')
+    const [app_id, setAppId] = useLocalStorageState(getAppId(), 'config.app_id')
```
ShuvoHabib, line 57: This, 'config.app_id' looks like a key inside an object. Should this be in string? Or can we rename the argument?

src/components/hooks/use-website-status.js
```diff
@@ -0,0 +1,39 @@
+import { useState, useEffect } from 'react'
+import { useCookieState } from './use-cookie-state'
+import { BinarySocketBase } from 'common/websocket/socket_base'
+import { getDateFromToday } from 'common/utility'
+
+const WEBSITE_STATUS_COUNTRY_KEY = 'website_status'
+
+export const useWebsiteStatus = () => {
+    const [website_status, setWebsiteStatus] = useCookieState(null, WEBSITE_STATUS_COUNTRY_KEY, {
+        expires: getDateFromToday(7),
```
ShuvoHabib, line 10: Perhaps we can use a variable the Number.

src/pages/endpoint/index.js
sean-binary: Don't think so as this page is only for us 😂

src/pages/endpoint/index.js
sean-binary: Don't think so as this page is only for us 😂

src/common/websocket/config.js
```diff
@@ -7,7 +7,9 @@
  *
  */
 import { isBrowser } from '../utility'
-const domain_config = {
+import { default_server_url } from '../constants'
+
+export const domain_config = {
     production: [
         {
             hostname: 'deriv.com',
```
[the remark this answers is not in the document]
mustofa-binary: hostname from constants has a logic to differentiate .com and .me, this config is to separate between the two

src/components/hooks/use-website-status.js
mustofa-binary: this is backend API convention, any api call will have the value 1, e.g payment_methods: 1, active_symbols: 1, etc.,

src/pages/endpoint/index.js
mustofa-binary: this page is only meant for us, no translation needed here

src/pages/endpoint/index.js
mustofa-binary: its not an object, its a name convention used by backend and qa to set app_id and server_url, changing this will also require them to change their automation test

src/components/hooks/use-website-status.js
mustofa-binary: did you mean 7?

ShuvoHabib (approved):
LGTM ~

src/common/utility.js
```diff
@@ -179,4 +179,23 @@ export const getLocalizedUrl = (path, is_index, to) => `/${path}${is_index ? ``
...
+}
+
+export const getDateFromToday = (num_of_days) => {
+    const today = new Date()
+    const end_date = new Date(today.getFullYear(), today.getMonth(), today.getDate() + num_of_days)
+
+    return end_date
+}
+
+export const isNullUndefined = (value) => value === null || value === undefined
```
yashim-deriv, line 191:
```suggestion
export const isNullUndefined = (obj) => obj === null || typeof obj === "undefined"
```
Using typeof in this case is better as it is guaranteed to return the "undefined" type if the object is uninitialized. Minor rename of params as we usually check objects for this case

src/components/hooks/use-cookie-state.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const useCookieState = (defaultValue, key, options) => {
+    const cookie_state = new CookieStorage(key)
+    const [value, setValue] = useState(() => {
+        const sticky_value = cookie_state.get(key)
+        const result = sticky_value
+            ? isJSONString(sticky_value)
+                ? JSON.parse(sticky_value)
+                : sticky_value
+            : defaultValue
```
yashim-deriv, line 13:
```suggestion
        const result = isJSONString(sticky_value) ? JSON.parse(sticky_value) : defaultValue
```
🚧 Comment #1
Do we need to check for falsy here if we already check it in `isJSONString` function? I think we should do all the parsing in a method that accepts two arguments, maybe something like:

`parseJSONString(value, default_value)`

parseJSON if succeed return parsed json else return `default_value`

src/common/utility.js
```diff
@@ -179,4 +179,23 @@ export const getLocalizedUrl = (path, is_index, to) => `/${path}${is_index ? ``
...
+
+export const isObject = (value) => typeof value === 'object'
+
+export const isJSONString = (value) => {
```
yashim-deriv, line 195: Check comment below~

src/components/hooks/use-localstorage-state.js
```diff
@@ -0,0 +1,25 @@
...
+import { isBrowser, isNullUndefined, isJSONString } from 'common/utility'
+
+export const useLocalStorageState = (defaultValue, key) => {
+    const [value, setValue] = useState(() => {
+        const sticky_value = isBrowser() ? window.localStorage.getItem(key) : null
+        return sticky_value
+            ? isJSONString(sticky_value)
+                ? JSON.parse(sticky_value)
+                : sticky_value
+            : defaultValue
```
yashim-deriv, line 11: Same as above~

yashim-deriv (commented):
Nice PR~ would speed up dev. Checkout the comments I left 😄

src/components/hooks/use-cookie-state.js
mustofa-binary: the idea was to parse sticky value if it is available, if not then use default value, however, there's possibility that sticky value is not a JSON (normal string). its a good idea to use parseJSONString function however, not with 2 arguments for a fallback, something more isolated will be like

```
const parseJSONString = (value) => isJSONString(value) ? JSON.parse(value) : value

const result = sticky_value ? parseJSONString(sticky_value) : defaultValue
```

src/common/utility.js
mustofa-binary: i think its better to check isJSONString and parseJSON and make it as isolated as possible, so it can be used in any type of situation

src/common/utility.js
mustofa-binary: nice catch 👌

src/components/hooks/use-cookie-state.js
mustofa-binary: the reason why not 2 arguments:
1. that one function will have too many operations
2. when reading the name of parseJSONString, normally people will expect it to parse `a` value (one argument), its hard to realize it will have more than one argument unless if you look back again to the function definition

yashim-deriv: I see, that is a legit reasoning 👍🏼

yashim-deriv (approved):
LGTM~

suthesan-deriv (approved):

sean-binary (dismissed):

sean-binary (approved):
LGTM

PR #1734, opened 2021-05-21
Blue/green deployment on staging

afshinpaydar-binary:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,9 +153,12 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            echo $CA_CRT | base64 --decode > ca.crt
+            docker build -t ${DOCKHUB_ORGANISATION}/deriv-com:${CIRCLE_SHA1} -t ${DOCKHUB_ORGANISATION}/deriv-com:<< parameters.docker_image_tag >> .
+
             git clone https://github.com/binary-com/devops-ci-scripts
-            cd ./devops-ci-scripts/k8s-build_tools
-            chmod +x release.sh
+            cd devops-ci-scripts/k8s-build_tools
+            chmod +x /release.sh
             echo $CA_CRT | base64 --decode > ca.crt
             ./release.sh deriv-com ${CIRCLE_SHA1}
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,8 +153,11 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            git clone https://github.com/binary-com/devops-ci-scripts
+            cd ./devops-ci-scripts/k8s-build_tools
+            chmod +x release.sh
             echo $CA_CRT | base64 --decode > ca.crt
-            kubectl --server=${KUBE_SERVER} --certificate-authority=ca.crt --token=$SERVICEACCOUNT_TOKEN set image deployment/<< parameters.k8s_svc_name >> << parameters.k8s_svc_name >>=${DOCKHUB_ORGANISATION}/deriv-com:${CIRCLE_SHA1}
+            ./release.sh deriv-com ${CIRCLE_SHA1}
 
 jobs:
   test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,7 +158,6 @@ commands:
 
             git clone https://github.com/binary-com/devops-ci-scripts
             cd devops-ci-scripts/k8s-build_tools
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -153,8 +153,11 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            git clone https://github.com/binary-com/devops-ci-scripts
+            cd ./devops-ci-scripts/k8s-build_tools
+            chmod +x release.sh
             echo $CA_CRT | base64 --decode > ca.crt
-            kubectl --server=${KUBE_SERVER} --certificate-authority=ca.crt --token=$SERVICEACCOUNT_TOKEN set image deployment/<< parameters.k8s_svc_name >> << parameters.k8s_svc_name >>=${DOCKHUB_ORGANISATION}/deriv-com:${CIRCLE_SHA1}
+            ./release.sh deriv-com ${CIRCLE_SHA1}
```
4p00rv, line 160: where are we specifying the namespace for new-staging?

.circleci/config.yml
```diff
@@ -153,8 +153,11 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            git clone https://github.com/binary-com/devops-ci-scripts
+            cd ./devops-ci-scripts/k8s-build_tools
+            chmod +x release.sh
```
gareth-deriv, line 158: Nitpick: This shouldn't be here - should commit it with +x in the devops-ci-scripts
You can merge and fix later though.

.circleci/config.yml
afshinpaydar-binary: In the Circleci environment variables

.circleci/config.yml
```diff
@@ -153,8 +153,11 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            git clone https://github.com/binary-com/devops-ci-scripts
```
gareth-deriv, line 156: Nitpick: In the future, we should instead have this in the image we're running in circle ci.
Whenever we do anything like this in the middle of a build, we're stalling circle ci, and delaying every build we run.

That's a note for the future though.

.circleci/config.yml
```diff
@@ -153,8 +153,11 @@ commands:
...
+            cd ./devops-ci-scripts/k8s-build_tools
+            chmod +x release.sh
             echo $CA_CRT | base64 --decode > ca.crt
-            kubectl --server=${KUBE_SERVER} --certificate-authority=ca.crt --token=$SERVICEACCOUNT_TOKEN set image deployment/<< parameters.k8s_svc_name >> << parameters.k8s_svc_name >>=${DOCKHUB_ORGANISATION}/deriv-com:${CIRCLE_SHA1}
```
gareth-deriv, line 157: We should leave *BOTH* of these for now, the old method to the old staging, AND the new method.

Otherwise the old domain, and the old staging system, won't get updated with new releases, and the frontend team would be left high and dry.

.circleci/config.yml
```diff
@@ -153,8 +153,11 @@ commands:
       - run:
           name: Deploying to k8s cluster for service << parameters.k8s_svc_name >>
           command: |
+            git clone https://github.com/binary-com/devops-ci-scripts
+            cd ./devops-ci-scripts/k8s-build_tools
```
gareth-deriv, line 157: Nitpick - I don't think it's healthy to CD like this, and I'm not sure why we do.

gareth-deriv (dismissed):
Good work.

I've added some feedback

.circleci/config.yml
afshinpaydar-binary: Yes, I agree with that

.circleci/config.yml
afshinpaydar-binary: Definitely will fix that latter

.circleci/config.yml
afshinpaydar-binary: Done

.circleci/config.yml
afshinpaydar-binary: Done

4p00rv (dismissed):

4p00rv (approved):

PR #1737, opened 2021-05-24
Remove the old release and use Blue/Green deployment as default

afshinpaydar-binary:
Switch to blue/green Deployment as a default release strategy on staging & production workflows
Before removing the old resources on default k8s namespace we need to Merge this

Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,31 +93,19 @@ commands:
   docker_build:
     description: "Build Docker image"
     parameters:
-      docker_image_tag:
+      docker_latest_image_tag:
         type: string
         default: "latest-staging"
+      docker_image_tag:
+        type: string
+        default: ${CIRCLE_SHA1}
     steps:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -153,16 +150,13 @@ commands:
     parameters:
       k8s_svc_name:
         type: string
-        default: "staging-deriv-com"
+        default: "deriv-com-staging"
```
4p00rv, line 153: Both `k8s_deploy` and `k8s_deploy_staging` have the same set of commands we can remove the `k8s_deploy_staging` config block and re-use `k8s_deploy` with `deriv-com-staging` as parameter

4p00rv (approved):

PR #1754, opened 2021-05-28
Suthesh/be-language-3

suthesan-deriv:
Changes:

-   added useClientInformation hook 
-   language switching based on cookie

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/hooks/use-client-information.js
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react'
+import { getClientInformation, getDomain } from 'common/utility'
+
+export const useClientInformation = () => {
+
+    const [client_information, setClientInformation] = useState(false)
+    const [current_client_information, setCurrentClientInformation] = useState(false)
```
sean-binary, line 7: I think the naming for these are a little misleading since they're actually boolean values. Could we change them to prefix with `is...` instead?

suthesan-deriv: They're meant to be initialised with something falsy...empty objects should be better

src/components/localization/language-switcher.js
```diff
@@ -13,11 +15,24 @@ const disabled_lang = ['ach']
...
     React.useEffect(() => {
         setLanguage(i18n.language)
     }, [i18n.language])
 
+    React.useEffect(() => {
+        if (!Cookies.get('lang_is_fixed')) {
+            if (client_information.preferred_language) {
+                const lang = client_information.preferred_language.toLowerCase()
+                if (lang !== language) {
+                    const replaced_lang = lang.replace('_', '-')
```
sean-binary, line 29: Can you chain this on line 27 as well?

suthesan-deriv: I don't think I can chain this, language state has _ so i need to replace after the check

src/components/hooks/use-client-information.js
sean-binary: Ah, empty object is a truthy value, so that won't do. You can initialise it with `null` and also set the state to null in your useEffect and try that.

sean-binary (approved):

PR #1779, opened 2021-06-02
Update nginx Config to include cache headers

haneef-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


deriv.com.conf
```diff
@@ -42,10 +52,10 @@ server {
     }
 
     location /static {
-        add_header Cache-Control "public, max-age=31536000, immutable";
+        add_header Cache-Control $cache_header always;
```
gareth-deriv, line 55: I note `always` and I wonder about it, in light of [the documentation](https://nginx.org/en/docs/http/ngx_http_headers_module.html)
We're using our status map to decide what to do here, but there's other status codes we haven't accounted for,
and we'd be defaulting to caching them.

I actually wonder if the lack of `always` has been ignored by all of us on the existing error responses.

Maybe that's *why* this worked and why Gokula chose to do it this way?

deriv.com.conf
```diff
@@ -1,7 +1,15 @@
+map $status $cache_header {
+    default "public, max-age=7200, s-maxage=600, must-revalidate";
```
gareth-deriv, line 2: Note for me, and maybe others:
`must-revalidate` doesn't mean what I thought it did
Two resources i'm reading to understand this better:
* [mdn documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
* [this post](https://sookocheff.com/post/api/effective-caching/)

deriv.com.conf
```diff
@@ -1,7 +1,15 @@
...
+    404     "max-age=0, no-cache, no-store, must-revalidate";
+    403     "max-age=0, no-cache, no-store, must-revalidate";
+    500     "max-age=0, no-cache, no-store, must-revalidate";
+}
+
 server {
     listen 80;
     server_name  _;
-
+    add_header Cache-Control $cache_header always;
```
gareth-deriv, line 11: I *think* we said that we should not ever cache the http page itself, and this would change that.

I think if we change this, we need to understand it.

I'd like to see what the behaviour looks like, specifically:
* Is our requirement that our releases are immediately respected
* If so, I don't think that'll happen with this Cache-Control header

gareth-deriv (commented):
I've raised questions and doubts, which maybe everyone else is okay with.

Personally, I think we should discuss, probably including the original gang from the first incident ( @arun-binary and @4p00rv ) as well as anyone with skin in the game from Frontend on this.

haneef-deriv:
A new pull request will be created

PR #1832, opened 2021-06-15
Sean/all-videos-page

sean-binary:
Changes:

-   

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/blog/videos/_video-player.js
```diff
@@ -0,0 +1,91 @@
...
+    @media ${device.tablet} {
+        width: 90%;
+    }
+`
+
+const VideoPlayer = ({ video_src, closeVideo }) => {
+    const handleEscape = (e) => {
+        if (e.keyCode === 27) {
+            closeVideo()
+        }
```
yashim-deriv, line 50: I think we should use the `event.key` as `event.keyCode` is marked as deprecated. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

yashim-deriv (commented):
Part 1 lunch comments

arome-binary:
@sean-binary  at the bottom of the figma design we have the subscription component, is that excluded in this phase?

src/pages/blog/videos/_video-card.js
```diff
@@ -0,0 +1,158 @@
...
+    color: var(--color-orange-2);
+    margin: 0 8px 8px 0;
+`
+
+const PlayButtonOval = styled.div`
+    position: absolute;
+    z-index: 2;
+    top: 50%;
+    left: 50%;
+    transform: translateX(-50%) translateY(-50%);
```
njazuli-deriv, line 80:
```suggestion
    transform: translate(-50%, -50%);
```

src/pages/blog/videos/_video-card.js
```diff
@@ -0,0 +1,158 @@
...
+    opacity: 0.32;
+    border-radius: 50%;
+`
+
+const StyledTriangle = styled.img`
+    position: absolute;
+    z-index: 4;
+    top: 50%;
+    left: 50%;
+    transform: translateX(-40%) translateY(-50%);
```
njazuli-deriv, line 93:
```suggestion
    transform: translate(-40%, -50%);
```

njazuli-deriv: same goes here

src/pages/blog/videos/_video-card.js
njazuli-deriv: we can shortened this for transform

sean-binary:
> @sean-binary at the bottom of the figma design we have the subscription component, is that excluded in this phase?

It will be added in once Habib's PR is merged.

src/pages/blog/videos/_all-videos.js
```diff
@@ -0,0 +1,58 @@
...
+    grid-row-gap: 40px;
+    grid-column-gap: 24px;
+    grid-template-rows: auto;
+`
+
+const AllVideos = ({ video_data }) => {
+    const [show, setShow] = useState(false)
+
+    const handleCloseVideo = () => setShow(false)
+    const handleOpenVideo = () => setShow(true)
```
yashim-deriv, line 27:
```suggestion
```
since this seems to be only used in one place each, i think we should just pass the function directly to the props instead of declaring a function here

src/pages/blog/videos/_data.js
```diff
@@ -0,0 +1,92 @@
+export const video_data = [
+    {
+        image: 'https://source.unsplash.com/random/1',
+        video_duration: '1:20',
+        category: ['Multipliers', 'CFD', 'Stocks', 'Forex', 'Category', 'Sixth'],
+        title: 'How to trade multipliers on Deriv',
+        description:
+            'To help you begin, we will look at the basics of cryptocurrency: what it is, how it works, and how to start trading cryptos on Deriv.com.',
+        date: '12 June 2021',
+        id: 1,
```
yashim-deriv, line 10: This index is for the `key` child props, correct? we can remove this and probably use other unique props like the `image` object instead?

yashim-deriv (commented):
Some comments (are open for discussion)

src/pages/blog/videos/_data.js
seanho96: Sure, although this is temporary. Will be replaced with the actual data's UUID in the future.

yashim-deriv: Niceee then LGTM 👍🏼

kevinw-binary (approved):

PR #1971, opened 2021-08-04
Sean/cache public and cache directories during GH actions build

sean-binary:
Changes:

-   Cache public and cache directories during GH actions build for Gatsby incremental builds
-   Changes made to add a new step `cache gatsby` & also removing excess lint step in `production.yml`

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [x] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,111 +1,109 @@
 name: Release Production
 
 on:
-  push:
-    tags:
-      - 'production*'
-    paths-ignore:
-      - '**.md' 
+    push:
+        tags:
+            - 'production*'
+        paths-ignore:
+            - '**.md'
 
 env:
-  GATSBY_CPU_COUNT: 2
-  GATSBY_ENV: production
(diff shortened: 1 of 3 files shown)


sean-binary:
@afshinpaydar-binary We also actually need to upgrade to Node 14 as well. I could do it together in this PR. Are there any foreseeable issues from your end if we do that?

afshinpaydar-binary (approved):

afshinpaydar-binary:
> @afshinpaydar-binary We also actually need to upgrade to Node 14 as well. I could do it together in this PR. Are there any foreseeable issues from your end if we do that?

No, there isn't any issue on k8s with that. You can change it

mustofa-binary (approved):

PR #2054, opened 2021-08-27
nikolai.berestevich/feat/add subscribe-banner on single article page

nikolaiberestevich-deriv:
Changes:

-  pages/academy/blog/components/side-subscription-banner/_agreement-label.js  deleted because use the one from components/custom/_agreement-label
- side Subscribe banner is added to template page(Single Post page )

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ x ] Update feature
-   [ x ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/academy/blog/components/side-subscription-banner/index.js
```diff
@@ -346,11 +355,30 @@ const ArticleEmailBanner = () => {
...
-                                isChecked={is_checked}
-                                color="#C2C2C2"
-                                handleChangeCheckbox={handleChange}
-                            />
+                            <>
+                                <AgreementLabel
+                                    isChecked={is_checked}
+                                    handleChangeCheckbox={handleChange}
+                                    color="#C2C2C2"
+                                    link_text="Send me marketing materials too!"
```
mahdiaryayi-fs, line 363: doesn't this `link_text` value need to be `localize`d?

sean-binary (commented):
Rest LGTM 👍

src/pages/academy/blog/components/side-subscription-banner/index.js
nikolaiberestevich-deriv: Yes, because the text near the checkbox is not "I agree to the terms and conditions" (link_text - additional prop was made by about 1 or 2 weeks ago. ) Sean approved this refactoring of the component.

sean-binary: @mahdiaryayi-fs It gets passed to a LocalizedText component.

sean-binary (approved):

src/pages/academy/blog/components/side-subscription-banner/index.js
mahdiaryayi-fs: AFAIK, `localize` is not working by passing variable (`localize(variable_string)`) and string should be passed directly into it (localize('some text')) because it is done by scanning the source code at build time (not run time)
If you are sure, you are not breaking this rule, it is OK

sean-binary: Ah yes, good point. Totally missed out on that. @nikolaiberestevich-deriv Could we `localize` the strings passed in as well? As @mahdiaryayi-fs pointed out, the needed strings would not be extracted during build time.

nikolaiberestevich-deriv: I think it should work fine because on the bottom of the testlink's homepage it works(AgreementLabel uses default prop :  link_text = 'I agree to the <0>terms and conditions</0>',). But today I can't say 100%

sean-binary: @nikolaiberestevich-deriv I think @mahdiaryayi-fs is referring to when we extract the strings onto Crowdin, we won't be able to extract the string that is passed in to `link_text` during this extraction stage when we run `npm run translate:extract`.

mahdiaryayi-fs: @nikolaiberestevich-deriv That's buggy too! the default one should change to: ` link_text = localize('I agree to the <0>terms and conditions</0>')`

sean-binary (approved):

PR #2079, opened 2021-09-08
Kevin/blog-hide-eu-feature

kevinw-binary:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/academy/blog/index.js
+++ b/src/pages/academy/blog/index.js
@@ -53,7 +53,10 @@ const StyledHeader = styled(Header)`
 
 const ArticlesPage = ({ data }) => {
     const { is_eu_country } = React.useContext(DerivStore)
-    const article_data = is_eu_country ? data.directus.blog_eu : data.directus.blog
+    const article_data = is_eu_country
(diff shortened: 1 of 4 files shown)


src/pages/academy/blog/index.js
```diff
@@ -112,6 +114,34 @@ export const query = graphql`
...
+                    id
+                    tags_id {
+                        tag_name
+                    }
+                }
+                blog_title
+                blog_description
+                read_time_in_minutes
+                published_date
+            }
```
sean-binary, line 144: For this page and also the videos page, we don't want to do a double query here as we are not limiting the query to a certain number. Instead, we can just declare `article_data` as `data.directus.blog.filter()` for the EU view.

sean-binary (approved):
LGTM

PR #2167, opened 2021-10-01
Alamin/habib/career page

habib-deriv:
Changes:

-   Career Page

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/careers/_join_your_team.js
```diff
@@ -0,0 +1,219 @@
...
+
+const DiffHeader = styled(Header)`
+    margin-bottom: 5rem;
+    font-size: 48px;
+    line-height: 60px;
+
+    @media ${device.tablet} {
+        font-size: 28px;
+        line-height: 34px;
+    }
```
yashim-deriv, line 18:
```suggestion
```

These sizing rule can be omitted. We already have handling for this set of style. You can set it by passing the prop `type="heading-2"` to the `Header` component. It should handle for everything including for mobile 🙏🏼

yashim-deriv (commented):
Left a comment regarding the font sizing. There are quite a few places in the PR where `font-size` & `line-height` is manually defined. Check if we can handle it by passing props to the `Header` component instead of manually extending the styles 🙏🏼

sean-binary:
This is no longer relevant hence will close it.

PR #2232, opened 2021-11-01
nikolai.berestevich/fix/academy-article-template banner fix

nikolaiberestevich-deriv:
Changes:

-   fix is  made not to show external link pop-up for side and bottom banners ( in academy article template)

## Type of change

-   [ x ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/templates/article.js
+++ b/src/templates/article.js
@@ -59,7 +59,7 @@ const ArticlesTemplate = (props) => {
     const side_banner_data_details = {
         max_w_value: '328px',
         max_w_tablet: '320px',
-        isExternal: true,
+        isExternal: footer_banner_data?.cta_url.search(/deriv\.(com|me)/g) === -1 ? true : false,
         redirectLink: side_banner_data?.cta_url,
         imgSrcDesktop: side_banner_data?.banner_image?.imageFile,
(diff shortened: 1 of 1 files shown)


src/templates/article.js
```diff
@@ -59,7 +59,7 @@ const ArticlesTemplate = (props) => {
     const side_banner_data_details = {
         max_w_value: '328px',
         max_w_tablet: '320px',
-        isExternal: true,
+        isExternal: footer_banner_data?.cta_url.search(/deriv\.(com|me)/g) === -1 ? true : false,
```
sean-binary, line 62: AFAIK all side and footer banners are meant to redirect to our sites, so I suppose we don't even need to check the URL, as they might possibly link to future platforms that are not part of this URL.

nikolaiberestevich-deriv: I've found another links that is not to deriv.com...

sean-binary (approved):

PR #2234, opened 2021-11-01
Pavel-Latyshou/on Deriv.com for EU and non-EU there is difference in redirecting pop up to the trading platfrom

pavel-latyshou-deriv:
Changes:

-   ...

## Type of change

-   [x ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/localization/localized-link.js
```diff
@@ -186,12 +185,10 @@ const ExternalLink = ({
     type,
     ...props
 }) => {
-    const { is_eu_country } = useContext(DerivStore)
```
sean-binary, line 189: We still need the modal pop for EU when it is an external link (not deriv product links). You need to move smart trader link onto the array that doesn't have the EU modal.

sean-binary (changes requested):

src/components/localization/localized-link.js
sean-binary: Should this not be removed anymore then? @pavel-latyshou-deriv

src/components/localization/localized-link.js
```diff
@@ -155,6 +155,8 @@ const getURLFormat = (type, locale, to, affiliate_lang) => {
         return getDerivAppLocalizedURL(localized_link_url[type], locale, to)
     } else if (affiliate_links.includes(type)) {
         return `${localized_link_url[type]}?lang=${affiliate_lang}`
+    } else if (deriv_social_platforms.includes(type) && type === 'smart_trader') {
+        return `${localized_link_url[type]}/${getThaiExcludedLocale(locale)}/${to}.html`
```
sean-binary, line 159: The only change that should be required for the PR should be to add `&& !deriv_other_products.includes(type)` on the conditions for `show_modal` on line 192. `smart_trader` shouldn't be categorised under social platforms.

sean-binary (changes requested):

sean-binary (approved):

yashim-deriv (changes requested):
Is there a reason to remove the `is_eu_country` check completely? To fix the redirect modal, you can simply add `!deriv_other_products.includes(type)` to not show modal for that specific case

PR #2261, opened 2021-11-23
if regulatory document does not exist, 404 page redirects to the engl…

mohammad-rework:


src/pages/404.js
```diff
@@ -40,6 +40,15 @@ const ButtonWrapper = styled.div`
...
 const PageNotFound = () => {
+    React.useEffect(() => {
+        if (window.location.pathname.includes('/regulatory/kid')) {
+            navigate(
+                window.location.origin +
+                    window.location.pathname.slice(0, 16) +
+                    window.location.pathname.slice(19),
+            )
+        }
+    }, [])
```
sean-binary, line 51: Hey @mohammad-rework, I think the fix here for that missing link would be to provide the PDF for that language in the directory. For this method, the client would still see the 404 screen before being routed back to the English version.

sean-binary, line 51: Hey @mohammad-rework, I think the fix here for that missing link would be to provide the PDF for that language in the directory. For this method, the client would still see the 404 screen before being routed back to the English version.

sean-binary (changes requested):

PR #2262, opened 2021-11-24
suthesh/blog-geo-conditions

suthesan-deriv:
Changes: 

-   filter contents for uk, eu, row

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/academy/index.js
```diff
@@ -43,6 +43,58 @@ export const query = graphql`
                     description
                 }
             }
+            homepage_banners_eu: homepage_banners(
+                filter: {
+                    status: { _eq: "published" }
+                    _or: [
+                        { visibility: { _eq: "hide_for_uk" } }
+                        { visibility: { _eq: "show_for_all" } }
```
sean-binary, line 51: I think it's about time we move this entire graphql query into its own file 😅

sean-binary (dismissed):
LGTM

src/common/country-base.js
```diff
@@ -27,7 +27,6 @@ const eu_countries = [
     'se',
     'ie',
     'im',
-    'gb',
```
sean-binary, line 30: We can't remove this from here as most of the regulatory contents follow EU/UK views together and by removing this, a client will not see the EU view if they are coming from GB.

sean-binary (changes requested):
The recent/featured section is breaking for EU view. Please do check on the fallback logic.

sean-binary (dismissed):

sean-binary (approved):

PR #2279, opened 2021-11-26
Vijay GitHub action env variable

vijay-deriv:
fetch GIT_TAG_NAME


vijay-deriv:
@shudarshon-binary please review

.github/workflows/content.yml
```diff
@@ -0,0 +1,133 @@
+name: Release Production
+
+on:
+    push:
+        tags:
+            - 'production*'
+        paths-ignore:
+            - '**.md'
```
shudarshon-binary, line 8: we can not allow tag based push operation for this new workflow. then it would be same as production. since this action is going to start by some events then we need to use repository dispatch. link to documentation: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#repository_dispatch

.github/workflows/content.yml
```diff
@@ -0,0 +1,133 @@
...
+              run: |
+                  git fetch --prune --unshallow --tags
+                  echo "GIT_TAG_NAME=`echo $(git describe --tags --abbrev=0)`" >> $GITHUB_ENV
+                  docker build -t ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:latest -t ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:$GIT_TAG_NAME .
+
+            - name: Pushing Image to docker hub 🐳
+              run: |
+                  echo ${{ secrets.DOCKERHUB_PASSWORD }}| docker login -u ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+                  docker push  ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:latest
+                  docker push  ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:$GIT_TAG_NAME
```
shudarshon-binary, line 104: we need to remove docker build and push steps to dockerhub steps since it will overwrite latest docker image. we do need to the tag values though.

vijay-deriv:
let me know if we need to change the event type ( https://github.com/vijay-deriv/deriv-com/blob/vijay-github-action-env-variable/.github/workflows/content.yml#L5)
cc: @shudarshon-binary

shudarshon-binary (dismissed):

.github/workflows/content.yml
```diff
@@ -0,0 +1,123 @@
+name: Release Production
```
sean-binary, line 1: Could we name this `Release Academy Content`?

.github/workflows/content.yml
```diff
@@ -0,0 +1,123 @@
...
+
+on:
+   repository_dispatch:
+     types: [publish_blog]
```
sean-binary, line 5: Let's change this to `publish_academy`.

vijay-deriv:
updated workflow name and event type @sean-binary

sean-binary (approved):

PR #2302, opened 2021-12-07
Convert check email to ts

sean-binary:
Changes:

-   Convert check email page to ts

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


njazuli-deriv (approved):

src/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+    const content: any
```
khalid-deriv, line 2: If you wanna avoid using `any` you can declare the type like here (https://duncanleung.com/typescript-module-declearation-svg-img-assets/). But it's still is a string after being imported, so wouldn't be that useful for type checking. Not sure though if it's better than using `any` 🤷🏼

prince-deriv: In most of documentations even in official gatsby they just use `any` and like what you said @khalid-deriv  It's not that very useful even if we are able to add a proper type for it. I think we can keep it as `any`

khalid-deriv (approved):

PR #2308, opened 2021-12-08
Mohammad / section-container responsive

mohammad-hashemi-deriv:


src/components/containers/section-container.js
```diff
@@ -1,15 +1,34 @@
...
+    @media ${device.tablet} {
+        padding: ${(props) => props.p || '8rem 0'};
+    }
+    @media ${device.tabletL} {
+        height: ${(props) => props.height};
+        padding: ${(props) => props.p};
+        padding-top: ${(props) => props.pt};
+    }
+
+    ${responsiveStyles}
```
prince-deriv, line 31: I think we don't need to add `responsiveStyles` here, It is already inherited in the parent component `Box`. The `props.p` condition alone can suffice the requirement.

src/components/containers/section-container.js
```diff
@@ -1,15 +1,34 @@
...
     background-color: var(--color-${(props) => props.background || 'white'});
+
+    @media ${device.tablet} {
+        padding: ${(props) => props.p || '8rem 0'};
+    }
+    @media ${device.tabletL} {
+        height: ${(props) => props.height};
+        padding: ${(props) => props.p};
+        padding-top: ${(props) => props.pt};
+    }
```
sean-binary, line 29: We don't need this right @prince-deriv?

prince-deriv: Yes @sean-binary

src/components/containers/section-container.js
```diff
@@ -4,12 +4,14 @@ import { size } from 'themes/device'
 
 const hasResponsivePadding = (props) => {
     let has_padding = false
-    const possibleOptions = ['p', 'pt', 'pr', 'pb', 'pl']
+    const padding_options = ['p', 'pt', 'pr', 'pb', 'pl']
 
     Object.keys(size).forEach((device) => {
```
prince-deriv, line 9: Can you add a comment, so later it will be clear to us why are we doing this.

src/components/containers/section-container.js
```diff
@@ -1,11 +1,27 @@
...
+
+    return has_padding
+}
 
 const SectionContainer = styled(Box).attrs({
     as: 'section',
 })`
     width: 100%;
-    padding: ${(props) => props.padding || '8rem 0'};
+    padding: ${(props) => props.padding || hasResponsivePadding(props) || '8rem 0'};
```
sean-binary, line 24: Should `hasResponsivePadding(props)` have a higher priority over `padding` in this case?

PR #2324, opened 2021-12-14
Mohammad/ Add "See all videos" button

mohammad-rework:
- Add "See all videos" button to Desktop and Mobile layout in /academy page.
- Remove previous "See all videos" in Mobile layout to avoid duplicate buttons.

### DIFF
--- a/src/pages/academy/components/video-banner/_DBanner.js
+++ b/src/pages/academy/components/video-banner/_DBanner.js
@@ -23,7 +23,7 @@ const ParentWrapper = styled(Flex)`
     position: relative;
     background-size: cover;
     height: auto;
-    padding: 160px 0 80px;
+    padding: 160px 0 40px;
 
     @media ${device.tabletL} {
         /* prettier-ignore */
@@ -70,9 +70,7 @@ const StyledDot = styled.img`
     margin: 0 10px 4px;
 `
 const AllVideosButton = styled(LinkButton)`
-    margin: 50px auto -40px;
-    width: fit-content;
-    padding: 12px 48px;
+    margin: 50px auto auto;
 
     &:hover {
         cursor: pointer;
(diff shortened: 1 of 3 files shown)


src/pages/academy/components/video-banner/_DBanner.js
```diff
@@ -68,6 +69,19 @@ const StyledDot = styled.img`
     background: var(--color-grey-17);
     margin: 0 10px 4px;
 `
+const AllVideosButton = styled(LinkButton)`
+    margin: 50px auto -40px;
```
sean-binary, line 73: I see that there are negative margins here, we usually try to avoid this. 😁  Also, is this a new button configuration from what we usually have in our base component? We can have a discussion about this if it should be added to the base component.

src/pages/academy/components/video-banner/_DBanner.js
```diff
@@ -164,6 +176,9 @@ const Dbanner = ({ featured_video_list, non_featured_video_list }) => {
                         </Flex>
                     </Flex>
                     <VideoCarousel carousel_items={non_featured_video_list} />
+                    <AllVideosButton white="true" to="/academy/videos/">
+                        See all videos
+                    </AllVideosButton>
```
sean-binary, line 181: I don't think the `white` type is the one you should be using as the hover state is different from the `all articles` button. It should be exactly the same as that one.

mohammad-rework: > I don't think the `white` type is the one you should be using as the hover state is different from the `all articles` button. It should be exactly the same as that one.

I can add a new button in `button.js` and name it `light` for example. make it exactly as same as the button we have right now, but copy the hover of the button used for `all articles`. Also that hover is a bit dark, is it OK to make it like this:

sean-binary: Is it different from our current tertiary button? If not, we can use that. Otherwise, we can create a new type. But do check with the designer first on inconsistency if it is a new variant or a slight mistake.

suthesan-deriv (approved):

PR #2352, opened 2021-12-19
Prince/ Show desktop mobile replacement

prince-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/containers/visibility.tsx
```diff
@@ -0,0 +1,49 @@
...
+import styled from 'styled-components'
+import device, { size } from 'themes/device'
+
+const DesktopLayer = styled.div`
+    @media ${device.tabletS} {
+        display: none;
+    }
+`
+const MobileLayer = styled.div`
+    @media (min-width: ${size.tablet}px) {
```
sean-binary, line 15:
```suggestion
    @media (min-width: ${size.tablet}) {
```

src/components/containers/visibility.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { ReactElement, useEffect, useState } from 'react'
+import styled from 'styled-components'
+import device, { size } from 'themes/device'
+import { useBrowserResize } from 'components/hooks/use-browser-resize'
+
+type ResponsiveContainerProps = {
+    children: ReactElement
```
sean-binary, line 7: I think this should be ReactNode as we may want to pass children in as well when calling it in another component.

src/components/containers/visibility.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+`
+const MobileLayer = styled.div`
+    @media ${device.mobile} {
+        display: none;
+    }
+`
+
+const deviceRenderer = (): boolean => {
+    const [is_loaded, setIsLoaded] = useState<boolean>(false)
```
sean-binary, line 22:
```suggestion
    const [is_loaded, setLoaded] = useState(false)
```
- AFAIK, the convention is that we usually remove the `is` & `has` for the setter.
- We can let TS infer the type here.

src/components/containers/visibility.tsx
```diff
@@ -0,0 +1,47 @@
...
+const DesktopLayer = styled.div`
+    @media ${device.tablet} {
+        display: none;
+    }
+`
+const MobileLayer = styled.div`
+    @media ${device.mobile} {
+        display: none;
+    }
+`
```
sean-binary, line 19: Might be better to use the bp from `size` instead of `device` since we want to control min-width and also max-width. We can accept the props of `size` and then pass it in here and also the `useBrowserResize`. Rest looks good 👍

src/components/containers/visibility.tsx
prince-deriv: @sean-binary I was using `reactNode` but as I told you I'm still getting this: 

```
'children' is missing in props validation (eslint react/prop-types)
```

sean-binary: Oh right, okay then ignore this comment then.

src/components/containers/visibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+type ResponsiveContainerProps = {
+    children: ReactElement
+    breakpoint?: number
+}
+
+type LayerProps = {
+    breakpoint?: number
+}
+
+const DEFAULT_BREAKPOINT = size.tablet
```
sean-binary, line 15: I think we should make this tabletL since we break at `992` a lot.

src/themes/device.js
```diff
@@ -17,6 +17,7 @@ export const size = {
 }
 
 const device = {
+    mobile: `(min-width: ${size.tablet}px)`,
```
sean-binary, line 20: Should be able to remove this now.

sean-binary (approved):

sean-binary (dismissed):

PR #2356, opened 2021-12-20
Al-amin/[Mobile responsive] Not able to scroll down to login with existing account in EU

alamin-deriv:
Changes:

-   [Mobile responsive] Not able to scroll down to login with existing account in EU

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/_signup-new.js
+++ b/src/components/custom/_signup-new.js
@@ -162,6 +162,8 @@ const LoginText = styled(Text)`
     margin-top: 1.6rem;
 
     @media ${device.tabletL} {
+        margin-top: ${(props) => props.mt};
+        margin-bottom: ${(props) => props.mb};
         font-size: 2rem;
     }
 `
@@ -324,12 +326,7 @@ const SignupNew = ({
                     <SocialText>Apple</SocialText>
                 </SocialButton>
             </SocialWrapper>
-            <LoginText
-                tabletL={{
-                    mb: is_eu_country ? '100px' : '0',
-                    mt: '3.75rem !important',
-                }}
-            >
+            <LoginText mt="3.75rem" mb={is_eu_country ? '100px' : '0'}>
                 {localize('Already have an account?')}
                 <StyledLinkText
                     id="dm-new-login-button"
--- a/src/components/custom/_signup-new.js
+++ b/src/components/custom/_signup-new.js
@@ -162,10 +162,6 @@ const LoginText = styled(Text)`
     margin-top: 1.6rem;
 
     @media ${device.tabletL} {
-        margin-bottom: ${(props) => props.margin_bottom};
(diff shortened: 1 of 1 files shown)


prince-deriv (dismissed):

sean-binary:
> LGTM @alamin-deriv
> 
> @sean-binary this is one example of limitation we have where we need to add `!important` to override the css rule, but in reality this `margin-top: 3.75rem` should be prioritized because it's a rule for specific size but since that's how `styled-component` works, we need to change our main components to be able to override css rules in specific sizes by default.

If not mistaken, passing in the `mt` as a prop to the component rather than declaring it in the CSS template literal should solve that specificity issue.

prince-deriv:
> If not mistaken, passing in the `mt` as a prop to the component rather than declaring it in the CSS template literal should solve that specificity issue.

The `LoginText` contains a `margin-top` rule for desktop and because of how `styled-components` prioritize the css rules, It will always override the ones we pass in the props

```
const LoginText = styled(Text)`
    text-align: center;
    align-self: center;
    margin-top: 1.6rem;

    @media ${device.tabletL} {
        font-size: 2rem;
    }
```

alamin-deriv:
> > If not mistaken, passing in the `mt` as a prop to the component rather than declaring it in the CSS template literal should solve that specificity issue.
> 
> The `LoginText` contains a `margin-top` rule for desktop and because of how `styled-components` prioritize the css rules, It will always override the ones we pass in the props
> 
> ```
> const LoginText = styled(Text)`
>     text-align: center;
>     align-self: center;
>     margin-top: 1.6rem;
> 
>     @media ${device.tabletL} {
>         font-size: 2rem;
>     }
> ```

``

sean-binary:
> > If not mistaken, passing in the `mt` as a prop to the component rather than declaring it in the CSS template literal should solve that specificity issue.
> 
> The `LoginText` contains a `margin-top` rule for desktop and because of how `styled-components` prioritize the css rules, It will always override the ones we pass in the props
> 
> ```
> const LoginText = styled(Text)`
>     text-align: center;
>     align-self: center;
>     margin-top: 1.6rem;
> 
>     @media ${device.tabletL} {
>         font-size: 2rem;
>     }
> ```

Yep so for components where responsive styles are required, we can do as such:

```jsx
<LoginText mt="1.6rem" tablet={{ mt: "3.75rem" }}></LoginText>

```
instead of declaring it in the CSS template literal above.

sean-binary (approved):

PR #2368, opened 2021-12-22
khalid/add deriv.be domain

khalid-deriv:
Changes:

-   Added deriv.be to the list of supported domains
- Added an app ID for deriv.be

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/common/websocket/config.js
+++ b/src/common/websocket/config.js
@@ -18,7 +18,7 @@ export const domain_config = {
     production: [
         {
             hostname: 'deriv.com',
-            app_id: deriv_com_app_id, // 11780,
+            app_id: deriv_com_app_id,
         },
         {
             hostname: 'deriv.me',
--- a/src/common/websocket/config.js
+++ b/src/common/websocket/config.js
@@ -7,21 +7,26 @@
  *
  */
 import { isBrowser } from '../utility'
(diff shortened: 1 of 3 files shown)


habib-deriv (approved):
Looks good to me

src/common/websocket/config.js
```diff
@@ -19,6 +19,10 @@ export const domain_config = {
             hostname: 'deriv.me',
             app_id: 1411,
         },
+        {
+            hostname: 'deriv.be',
+            app_id: 30767,
+        },
```
sean-binary, line 25: Can we update this and the above to use the variables from `constants.js`? Also noticed that the one for deriv.com is different with the one in constants.js. Could you help to check on that @khalid-deriv?

khalid-deriv: Done!
And for deriv.com the second app ID also does the same thing, so I removed one for consistency (after checking with @habib-deriv)

habib-deriv (approved):

PR #2375, opened 2021-12-23
Mohammad/Call live chat after full window load.

mohammad-rework:
1. **Delayed** component created (It gets `children` and `delay` as props)
2. In layout.js, `Livechat` is passed inside the `Delayed` component with delay of 500ms.


ali-hosseini-fs:
I believe the problem with livechat is with it's bundle files that loads directly and as soon as app loads. IMO what we should do is a delay or even apply an event listener to wait for a user gesture to deliver the related bundle files.

PR #2407, opened 2021-12-29
Niloofar Sadeghi / Call the Live Chat only when user click the chat icon

niloo-fs:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [x] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/hooks/use-livechat.js
```diff
@@ -52,10 +53,11 @@ export const useLivechat = () => {
             clearInterval(cookie_interval)
             clearTimeout(script_timeout)
         }
-    }, [])
+    }, [runScripts])
```
prince-deriv, line 56: isn't `isBrowser` a dependency in this hook?

niloo-fs: no it is not, they only wrote a isBrowser condition inside the useEffect.

PR #2495, opened 2022-01-11
Update Dp2p payment method section

mitra-fs:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


mahdiyeh-fs (dismissed):

src/pages/payment-methods/_mobile-expanded-list.js
```diff
@@ -84,6 +86,9 @@ const StyledRefLink = styled(Flex)`
     @media ${device.mobileL} {
         justify-content: flex-start;
     }
+    @media ${device.tabletM} {
+        width: 65%;
+    }
```
suthesan-deriv, line 91: Can we can use Flex here and pass in most of these as props?

src/pages/payment-methods/index.js
```diff
@@ -271,9 +271,15 @@ const DisplayAccordianItem = ({ pd, crypto_config, locale }) => {
                                     </Th>
                                 )}
 
-                                <Th>
-                                    <BoldText>{localize('Reference')}</BoldText>
-                                </Th>
+                                {pd.is_dp2p ? (
+                                    <Th>
+                                        <BoldText>{localize('More info')}</BoldText>
```
habib-deriv, line 276: Put the logic here, that will reduce duplicating the whole table

src/pages/payment-methods/_payment-data.js
```diff
@@ -849,12 +849,8 @@ const payment_data = [
                 deposit_time: <Localize translate_text="500 USD" />,
                 withdrawal_time: <Localize translate_text="Max 2 hours" />,
                 reference_link: (
-                    <StyledRefLink
-                        to="/help-centre/deriv-p2p/#what-is-deriv-p2p"
-                        target="_blank"
-                        rel="noopener noreferrer"
-                    >
-                        Deriv P2P FAQ
+                    <StyledRefLink to="/p2p" target="_blank" rel="noopener noreferrer">
```
habib-deriv, line 852: Add a forward slash after the link

src/pages/payment-methods/_mobile-expanded-list.js
mitra-fs: @suthesan-deriv Hi , could you please elaborate more on this ? or any part of the code using it and I can refer to it? 


Also as another work around  I thought of removing the 50% width only In `StyledItemDiv` to let it be more responsive. like this 

Screen Shot 2022-01-13 at 10 32 36 AM

Please do let me know if there is a better way , thanks 🙏

src/themes/device.js
```diff
@@ -20,6 +20,7 @@ const device = {
     mobileS: `(max-width: ${size.mobileS}px)`,
     mobileM: `(max-width: ${size.mobileM}px)`,
     mobileL: `(max-width: ${size.mobileL}px)`,
+    tabletM: `(min-width: ${size.mobileL}px) and (max-width: ${size.tabletL}px)`,
```
sean-binary, line 23: Do we really need this? Or if it is just a special case, then you can use the hard breakpoints instead.

mitra-fs: Hi @sean-binary per this comment https://github.com/binary-com/deriv-com/pull/2495#discussion_r783680266 I gave , If I use that workaround provided there then this part won’t be needed and can remove it , just waiting to see if it's okay to make that change 🙏

ShuvoHabib (approved):

habib-deriv (approved):

src/pages/payment-methods/_mobile-expanded-list.js
```diff
@@ -236,8 +242,11 @@ const MobileExpandedList = ({ is_crypto, is_fiat_onramp, is_dp2p, locale, paymen
                     {/* reference row */}
                     <StyledRow jc="space-between" ai="center">
                         <StyledItemDiv>
-                            <Header type="subtitle-2">{localize('Reference')}</Header>
+                            <Header type="subtitle-2">
```
sean-binary, line 245:
```suggestion
                            <Header as="p" type="subtitle-2">
```

sean-binary: Please include the `as` prop for the Header component if you notice any missing, otherwise they will default to a `h1` element when in this case it shouldn't be.

src/themes/device.js
sean-binary: Yep please do it in that component level and remove it here 🙏  But as @suthesan-deriv was suggesting, can we not do it with `tabletL`/`tablet`/`tabletS` props?

PR #2566, opened 2022-01-26
Mitra/Change Learn more about Deriv P2P to Learn more

mitra-fs:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/payment-methods/_mobile-expanded-list.js
```diff
@@ -116,7 +110,9 @@ const MobileExpandedList = ({ is_crypto, is_fiat_onramp, is_dp2p, locale, paymen
                     {/* first row */}
                     <StyledRow jc="space-between" ai="center">
                         <StyledItemDiv>
-                            <Header type="subtitle-2">{localize('Currencies')}</Header>
+                            <Header as="p" type="subtitle-2">
```
mitra-fs, line 113: Added `as "p"` tag per comment in pervious PR which overlooked 🙏 

https://github.com/binary-com/deriv-com/pull/2495/files#r785597175

src/pages/payment-methods/_mobile-expanded-list.js
```diff
@@ -86,10 +84,6 @@ const StyledRefLink = styled(Flex)`
     @media ${device.mobileL} {
         justify-content: flex-start;
     }
-
-    @media ${device.tabletM} {
-        width: 65%;
```
mitra-fs, line 91: Removing this part as we needed it before because the sentence was longer and would be broken in tablet mode.

src/pages/payment-methods/_mobile-expanded-list.js
sara-fs: Why don't we use <text> instead? what's the reason to add `p` ?

mitra-fs: > Why don't we use `<text>` instead? what's the reason to add `p` ?

Hi @sara-fs many thanks for reviewing , as I checked this with Sean seems `<text>` is now deprecated as `Header` component has the correct size mapping from the design team. 🙏

njazuli-deriv (dismissed):

njazuli-deriv (approved):

PR #2613, opened 2022-02-07
Niloofar Sadeghi / Fix hamburger-menu bug in partner nav

niloo-fs:
Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/nav.js
+++ b/src/components/layout/nav.js
@@ -1089,14 +1089,12 @@ export const NavPartners = ({ no_login_signup }) => {
                                 src={Close}
                                 alt="close menu 2"
                                 onClick={closeOffCanvasMenu}
-                                width="16px"
                             />
                         ) : (
                             <HamburgerMenuPartners
                                 src={Hamburger}
                                 alt="hamburger menu2"
                                 onClick={openOffCanvasMenu}
-                                width="16px"
                             />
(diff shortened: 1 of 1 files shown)


salarhfz-fs (approved):

src/components/layout/nav.js
```diff
@@ -1089,14 +1089,12 @@ export const NavPartners = ({ no_login_signup }) => {
                                 src={Close}
                                 alt="close menu 2"
                                 onClick={closeOffCanvasMenu}
-                                width="16px"
```
sean-binary, line 1092: Instead of deleting this, you can specify it like `width={16}` or have the width set in `CloseMenuPartners` styles above itself. This is was a breaking change when we upgraded to Preact and it doesn't accept pixels for inline width and height.

niloo-fs: There is not any difference between passing 16 and nothing, because of that I removed it

sean-binary (approved):

PR #2616, opened 2022-02-07
MohammadH/ In all-articles page scroll to the spot where the user was

mohammad-rework:
…on the post card when they come back to all-articles page

Changes:

-   In “All articles” when we select some article and land on the page, and click “back” to the [https://deriv.com/academy/blog/](url) page load from the top. Need to ensure the user can land back to the same spot they left.

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


alamin-deriv (approved):

src/pages/academy/blog/index.tsx
```diff
@@ -69,6 +69,38 @@ const ArticlesPage = ({ data }) => {
         og_description:
             'Educational content at your fingertips – everything you need to know to start trading or upgrade your trading skills.',
     }
+
+    const [scrollPosition, setScrollPosition] = React.useState(0)
```
prince-deriv, line 73:
```suggestion
    const [scroll_position, setScrollPosition] = React.useState(0)
```

src/pages/academy/blog/index.tsx
```diff
@@ -69,6 +69,38 @@ const ArticlesPage = ({ data }) => {
...
+    const handleScroll = () => {
+        const position = window.pageYOffset
+        setScrollPosition(position)
+    }
+
+    React.useEffect(() => {
```
prince-deriv, line 79: import `useEffect` from react 

```suggestion
   useEffect(() => {
```

src/pages/academy/blog/index.tsx
```diff
@@ -69,6 +69,38 @@ const ArticlesPage = ({ data }) => {
         og_description:
             'Educational content at your fingertips – everything you need to know to start trading or upgrade your trading skills.',
     }
+
+    const [scrollPosition, setScrollPosition] = React.useState(0)
+    const handleScroll = () => {
+        const position = window.pageYOffset
+        setScrollPosition(position)
```
salarhfz-fs, line 76:
```suggestion
        setScrollPosition(window.pageYOffset)
```

src/pages/academy/blog/index.tsx
```diff
@@ -69,6 +69,38 @@ const ArticlesPage = ({ data }) => {
...
+            window.removeEventListener('scroll', handleScroll)
+        }
+    }, [])
+
+    React.useEffect(() => {
+        scrollPosition && localStorage.setItem('scroll', scrollPosition.toString())
+    }, [scrollPosition])
+
+    React.useEffect(() => {
+        const prevScroll = parseInt(localStorage.getItem('scroll'))
```
salarhfz-fs, line 91: I think if `localStorage.getItem('scroll')` return `null` then becomes `NaN`. What do you think?

mohammad-rework: You are right but I think it never returns `null` due to `useEffect` on line 79. What do you think?

salarhfz-fs: I think there are chances. In the first page load for example.
We can simple prevent hat by adding `prevScroll !== NaN` to the condition.
BTW I'm not sure 100%, just a suggestion.

salarhfz-fs (approved):

PR #2620, opened 2022-02-09
Academy page logo update

ashraf-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


ashraf-deriv:
I am removing academy_logo flag references from all referencing pages since it is not using now.

salarhfz-fs (approved):

prince-deriv (approved):

sean-binary:
I believe we can just use the `Nav` instead of `NavAcademy` in layout, since there isn't any differences now. @ashraf-deriv

ashraf-deriv:
> I believe we can just use the `Nav` instead of `NavAcademy` in layout, since there isn't any differences now. @ashraf-deriv

Thanks Sean for the comment, Looks like It is possible, If I am not wrong we still have to handle the visibility of nav bar while scrolling the page specifically for academy page.

sean-binary:
Alright let's keep it separate for separation of concerns, as I overlooked that you would also need to handle `SearchBanner` as well.

ashraf-deriv:
> Alright let's keep it separate for separation of concerns, as I overlooked that you would also need to handle `SearchBanner` as well.

Yes that is right, we can keep it like that for now. Thanks for the review @sean-binary

src/pages/academy/components/_search-banner.tsx
```diff
@@ -13,7 +13,7 @@ import { useBrowserResize } from 'components/hooks/use-browser-resize'
 import { slugify, isBrowser } from 'common/utility'
 import { DerivStore } from 'store'
 import device from 'themes/device'
-import AcademyLogo from 'images/svg/blog/academy-logo.svg'
+import AcademyLogo from 'images/svg/blog/academy-text.svg'
```
sean-binary, line 16: Could you also please remove the original `academy-logo.svg` if it is not in used anymore?

ashraf-deriv: @sean Ok noted, will update that. Thank you

sean-binary (approved):

PR #2637, opened 2022-02-11
Vijayasree/ refactor: update who-we-are to new-show-component

vijayasree-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/who-we-are/_OurOffices.tsx
+++ b/src/pages/who-we-are/_OurOffices.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 import { graphql, useStaticQuery } from 'gatsby'
 import { desktop_pins, mobile_pins, MapPinType } from './_data'
 import { localize, LocalizedLink } from 'components/localization'
-import { SectionContainer, CssGrid, Show, Flex } from 'components/containers'
+import { SectionContainer, CssGrid, Desktop, Mobile, Flex } from 'components/containers'
 import { Header, Text, BackgroundImage } from 'components/elements'
 import device from 'themes/device'
 import { ReactComponent as Pin } from 'images/svg/who-we-are/pin.svg'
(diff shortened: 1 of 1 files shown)


src/pages/who-we-are/_OurOffices.tsx
```diff
@@ -173,7 +173,7 @@ const OurOffices = () => {
             </StyledHeader>
 
             <Flex>
-                <Show.Desktop max_width="tabletL">
+                <Desktop max_width="tabletL">
```
sean-binary, line 176: I don't think `Desktop` and `Mobile` uses `max_width` props. Please take a look at the `visibility.tsx` file.

sean-binary (changes requested):

src/pages/who-we-are/_OurOffices.tsx
vijayasree-deriv: The Desktop component renders a DesktopLayer component which has a media query where max_width uses a breakpoint value. (Similarly with Mobile component , MobileLayer and min_width.)

sean-binary: Yep but the prop should be `breakpoint` in that case. This is still working as the default of the `Desktop` component is `tabletL`.

vijayasree-deriv: yes you are right, I will make the necessary changes and update.

sean-binary (approved):

sean-binary (approved):

PR #2644, opened 2022-02-14
feat: add graphql codegen & refactor academy pages

sean-binary:
Changes:

-   Add graphql types codegen
-   Refactor academy pages to TS

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


njazuli-deriv (dismissed):
LGTM..all pages works fine..

maryia-deriv:
I would like to suggest looking into the [5 Code Smells](https://sonarcloud.io/project/issues?id=binary-com_deriv-com&pullRequest=2644&resolved=false&types=CODE_SMELL) discovered by SonarQube (TypeScript) service as a result of TS migration for the sake of code health even though the code smells are outside the lines you explicitly modified. 😊

src/components/elements/query-image.tsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
 import { GatsbyImage, getImage } from 'gatsby-plugin-image'
-import type { ImageDataLike } from 'gatsby-plugin-image'
+import type { ImageDataLike, IGatsbyImageData } from 'gatsby-plugin-image'
 
 type QueryImageProps = {
     alt: string
     className?: string
```
habib-deriv, line 8: Should not this className, height and width be always string ?

src/pages/academy/common/_styles.ts
```diff
@@ -22,7 +22,12 @@ export const Container = styled(Flex)`
     }
 `
 
-export const VideoGrid = styled.div`
+type VideoGridProps = {
+    m?: string
```
habib-deriv, line 26: Can we use a descriptive prop name here?

src/pages/academy/common/_styles.ts
```diff
@@ -34,7 +39,19 @@ export const VideoGrid = styled.div`
     justify-content: center;
 `
 
-export const StandardImgWrapper = styled.div`
+type StandardImgWrapperProps = {
+    width?: string
+    height?: string
+    br?: string
```
habib-deriv, line 45: Same for this

src/pages/academy/videos/_all-videos.tsx
```diff
@@ -1,15 +1,19 @@
...
 import { Text, LocalizedLinkText } from 'components/elements'
 import RightArrow from 'images/svg/tools/black-right-arrow.svg'
 
-const AllVideos = ({ video_data }) => {
+type AllVideosProps = {
+    video_data: VideoDataType
+}
+
+const AllVideos = ({ video_data }: AllVideosProps) => {
     const [show, setShow] = useState(false)
```
habib-deriv, line 17: Instead of 'false', should we use, ''; As overflow to 'false' doesn't sound right ?

src/pages/academy/_video-banner.tsx
```diff
@@ -0,0 +1,27 @@
...
+    non_featured_video_list_data: NonFeaturedVideoListDataType
+}
+
+const VideoBanner = ({
+    featured_video_list_data,
+    non_featured_video_list_data,
+}: VideoBannerProps) => {
+    return (
+        <Flex>
+            {non_featured_video_list_data && non_featured_video_list_data.length && (
```
habib-deriv, line 17: Let's use the shorthand operator, '?' like evrywhere else

src/components/elements/query-image.tsx
sean-binary: Sorry I don't quite understand your question. Yep it is now currently set to only be string.

habib-deriv: I meant as it says with the '?', it can be undefined as well. Do we have a chance this className can be undefined in this component?, If not, we can remove this otherwise we can keep this.

src/pages/academy/videos/_all-videos.tsx
sean-binary: The `show` is actually meant to toggle on/off the video player. The overflow is just styled accordingly dependent on that state. I think this is fine.

src/pages/academy/_video-banner.tsx
sean-binary: Sorry, how is the ternary operator a shorthand in this case?

sean-binary:
Thanks @maryia-deriv for pointing that out. I've addressed most of it but I think the one regarding the cognitive complexity is fine and the other regarding the unnecessary `await` is incorrect.

src/components/elements/query-image.tsx
sean-binary: Yes the `?` is present as these are optional props passed to QueryImage. They may not be used at times.

njazuli-deriv (approved):
LGTM

PR #2655, opened 2022-02-16
Niloofar Sadeghi / Refactor navigation component 

niloo-fs:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/layout/nav/nav-careers.tsx
```diff
@@ -116,33 +133,21 @@ export const NavCareers = ({ is_besquare }) => {
...
-                                active={current_page === 'besquare'}
-                                activeClassName="active"
-                                to="/careers/besquare/"
-                                aria-label={'BeSquare'}
-                                partiallyActive={true}
-                            >
-                                BESQUARE
-                            </CareerLink>
+
+                            {links.map((item) => {
```
ali-hosseini-fs, line 137: don't need to use return 
you can simply remove `{ return (` and just write `(item) => <CareerLink ...`

src/components/layout/nav/nav-desktop.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+    const setDropdownRef = (new_ref) => setActiveDropdownRef(new_ref)
+
+    useOutsideClick(navigation_bar_ref, () => setActiveDropdown(''), active_dropdown_ref)
+
+    useEffect(() => {
+        setMounted(true)
+        document.addEventListener('scroll', buttonHandleScroll, { passive: true })
+        return () => document.removeEventListener('scroll', buttonHandleScroll)
+    }, [])
```
ali-hosseini-fs, line 122: this will throws exhaustive deps warning when checking code with eslint
because you are using setMounted inside useEffect but didn't add it to deps array

src/components/layout/nav/nav-desktop.tsx
```diff
@@ -0,0 +1,239 @@
...
+                        <img src={LogoCombinedShape} alt={'logo combined shape'} />
+                    )}
+                </NavLeftMain>
+
+                <NavCenter ref={navigation_bar_ref}>
+                    {tabs.map((item) => {
+                        return (
+                            <NavLink key={`tab-${item.id}`} onClick={item.onClick}>
+                                <StyledButton
+                                    aria-label={localize(item.title)}
```
ali-hosseini-fs, line 192: we can't pass variables to localize() since we have file readers for translation generator.

src/components/layout/nav/nav-jump-indices.tsx
```diff
@@ -97,13 +94,9 @@ const JumpIndiceNav = ({ is_ppc }) => {
                     </LogoWrapper>
                 </ContentContainer>
             </Section>
-            <CFDWarning is_ppc={is_ppc} />
+            <CFDWarning {...props} />
```
ali-hosseini-fs, line 97: the spread props pattern will make the code unclear a little bit. why do we need this? this one is not recommended

src/components/layout/nav/nav-partner.tsx
```diff
@@ -0,0 +1,371 @@
...
+
+    return (
+        <DerivHomeWrapper>
+            <HomeContainer justify="space-between">
+                <StyledContainer justify="flex-start">
+                    {links.map((item) => {
+                        return (
+                            <HomeLink key={item.id} to={item.to}>
+                                <Text color="grey-19" size="var(--text-size-xxs)">
+                                    {localize(item.title)}
```
ali-hosseini-fs, line 191: passing variables to localize will break the translations

src/components/layout/nav/nav-sticky.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+type NavStickyProps = {
+    is_ppc: boolean
+    hide_signup_login?: boolean
+    no_language?: boolean
+}
+
+const NavSticky = ({ is_ppc, hide_signup_login, no_language }: NavStickyProps) => {
+    const [is_logged_in, setLoggedIn] = useState(true)
+    const [prevScrollPos, setPrevScrollPos] = useState(0)
```
ali-hosseini-fs, line 36: our naming convention for states are like the one in the upper line
prev_scroll_pos 
I know this is odd :-D

ali-hosseini-fs:
home page nav bar style is different.
please compare it with deriv.com

ali-hosseini-fs:
on `/partners/` page go to mobile view. then try to click on burger menu
the will hide after that

src/components/layout/nav/models/nav-types.ts
```diff
@@ -0,0 +1,10 @@
+export type NavTypes = {
+    base?: string
```
habib-deriv, line 2: Why is not the types fixed?

yashim-deriv (changes requested):
The CSS for different navigation bars can actually be shared. Try extracting that out first. Though the splitting of different util function is good, the whole idea of the refactor is to use only one base Nav for the other Nav to put contents in.

src/components/layout/nav/models/nav-types.ts
niloo-fs: these properties are the same in different nav component, because of that I use global types for them, but when I checked layout we didn't use all the properties so I did that, but I'll remove the extra property in my next changes

src/components/layout/nav/nav-desktop.tsx
```diff
@@ -0,0 +1,239 @@
...
+        document.addEventListener('scroll', buttonHandleScroll, { passive: true })
+        return () => document.removeEventListener('scroll', buttonHandleScroll)
+    }, [])
+
+    const tabs = [
+        {
+            id: 1,
+            title: 'Trade',
+            active: checkActive('trade'),
+            onClick: (e) => handleLinkClick('trade', e.target),
```
habib-deriv, line 129: Instead of passing onClick to every item, we can just use the handleLinkClick method. We can just pass the lowercase title via argument

src/components/layout/nav/nav-interim.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <NavInterimContainer>
+            <Container jc="space-between" p="2.4rem 0">
+                <Flex ai="center" jc="flex-start">
+                    <DesktopWrapper>
+                        <StyledLogo to={`/interim/${interim_type}`} aria-label={localize('Home')}>
+                            <Flex ai="center">
+                                <img src={Logo} alt="logo" width="190" height="27" />
+                                <img
+                                    src={LogoCombinedShape}
+                                    alt="logo combined shape desktop"
```
habib-deriv, line 68: Let's change alt text to something better

src/components/layout/nav/nav-interim.tsx
```diff
@@ -0,0 +1,109 @@
...
+                            to={`/interim/${interim_type}`}
+                            aria-label={localize('Home')}
+                        >
+                            <Flex>
+                                <img src={LogoOnly} alt="logo only 2" width="115" height="27" />
+                                <LogoDescription ai="center">
+                                    <Line />
+                                    <img
+                                        src={LogoCombinedShape}
+                                        alt="logo combined shape mobile"
```
habib-deriv, line 87: Same here

src/components/layout/nav/nav-mobile.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    return (
+        <Wrapper width="95%">
+            {is_canvas_menu_open ? (
+                <CloseMenu src={Close} alt="close menu" onClick={closeOffCanvasMenu} />
+            ) : (
+                <HamburgerMenu src={Hamburger} alt="hamburger" onClick={openOffCanvasMenu} />
+            )}
+
+            <LogoLinkMobileMain to="/" aria-label={localize('Home')}>
```
habib-deriv, line 82: I don't think we need localization for aria-label

src/components/layout/nav/nav-partner.tsx
```diff
@@ -0,0 +1,371 @@
...
+                                <img src={LogoPartner} alt="logo partner" />
+                            </NavLogoLink>
+                        </NavLeftPartners>
+                        <StyledNavCenter>
+                            <NavLink>
+                                <StyledLink
+                                    active={current_page === 'affiliate'}
+                                    activeClassName="active"
+                                    to="/partners/affiliate-ib/"
+                                    aria-label={localize('Affiliates and IBs')}
```
habib-deriv, line 255: Same here as before

src/components/layout/nav/nav-static.tsx
```diff
@@ -0,0 +1,43 @@
...
+type NavStaticProps = {
+    is_ppc: boolean
+}
+
+const NavStatic = (props: NavStaticProps) => (
+    <>
+        <StaticWrapper>
+            <LogoLink mw="31rem" to="/" aria-label={localize('Home')}>
+                <Flex ai="center">
+                    <img src={LogoOnly} alt="logo only nav static" width={160} height={27} />
```
habib-deriv, line 28: Same as before for alt

src/components/layout/nav/nav-academy.tsx
```diff
@@ -0,0 +1,37 @@
...
+        setPrevScrollPosition(current_scroll_position)
+        setVisible(current_scroll_position > 72)
+    }
+
+    useEffect(() => {
+        is_mounted && scrollHandler()
+    }, [is_mounted])
+
+    useEffect(() => {
+        window.addEventListener('scroll', scrollHandler)
```
yashim-deriv, line 25:
```suggestion
        window.addEventListener('scroll', scrollHandler)
```
🙏🏼  please add passive scroll handlers

src/components/layout/nav/nav-sticky.tsx
```diff
@@ -0,0 +1,48 @@
...
+        setVisible(
+            (prev_scroll_position > current_scroll_position &&
+                prev_scroll_position - current_scroll_position > 70) ||
+                current_scroll_position < 10,
+        )
+        setPrevScrollPosition(current_scroll_position)
+    }, [])
+
+    useEffect(() => {
+        window.addEventListener('scroll', backgroundHandler)
```
yashim-deriv, line 28:
```suggestion
        window.addEventListener('scroll', backgroundHandler)
```
Please add passive handlers here as well

yashim-deriv (commented):
Some comments for now

src/components/layout/nav/nav-interim.tsx
```diff
@@ -0,0 +1,84 @@
...
+        width: 100%;
+    }
+`
+
+const NavInterim = ({ interim_type }: NavInterimProps) => (
+    <NavTemplate nav_height="8.6rem">
+        <Container jc="space-between" p="2.4rem 0">
+            <Flex ai="center" jc="flex-start">
+                <DesktopWrapper>
+                    <LogoWrapper to={`/interim/${interim_type}`} aria-label="Home">
```
habib-deriv, line 53: Add a slash in the end of the url for all

src/components/layout/nav/util/useAuth.tsx
```diff
@@ -0,0 +1,16 @@
...
+    useEffect(() => {
+        setLoggedIn(isLoggedIn())
+        const checkCookieChange = setInterval(() => setLoggedIn(isLoggedIn()), 800)
+        return () => clearInterval(checkCookieChange)
+    }, [])
+
+    return [is_logged_in]
+}
+
+export default useAuth
```
sean-binary, line 16: I think this is general enough to be together in the `hooks` folder. Please name it in snake_case for hooks, `use-auth.tsx`.

niloo-fs: sure @sean-binary I'll move this there

src/components/layout/nav/styles/nav-styles.ts
```diff
@@ -0,0 +1,203 @@
...
+    .fresnel-between-start-tabletL {
+        display: none;
+    }
+    @media ${device.tabletL} {
+        .fresnel-between-start-tabletL {
+            display: flex;
+            width: 100%;
+            align-items: center;
+        }
+    }
```
sean-binary, line 202: Could you help to verify if we still use `Show` components in our Navs? If not, we can remove this.

src/pages/index.tsx
```diff
@@ -13,7 +13,7 @@ const Home = () => {
     useOpenLiveChat()
 
     return (
-        <Layout is_nav_transparent margin_top="0">
+        <Layout is_nav_transparent type="sticky" margin_top="0">
```
sean-binary, line 16: Does `is_nav_transparent` and `type="sticky"` doing the same thing?

niloo-fs: thank u @sean-binary I didn't see this before. cause I just updated my branch with master. I'll fix that

src/components/layout/nav/styles/nav-styles.ts
niloo-fs: No, we don't use it here anymore

src/components/layout/nav/components/partner/main-nav.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <Wrapper>
+            <StyledContainer justify="space-between">
+                <Navigation justify="flex-start">
+                    {links.map((item) => (
+                        <Link key={item.id} to={item.to}>
+                            <Header weight="normal" color="grey-19" type="paragraph-2">
+                                {item.title}
+                            </Header>
+                        </Link>
+                    ))}
```
yashim-deriv, line 61:
```suggestion
                    {links.map(({ id, to, title }) => (
                        <Link key={id} to={to}>
                            <Header weight="normal" color="grey-19" type="paragraph-2">
                                {title}
                            </Header>
                        </Link>
                    ))}
```
Cleaner just to de-structure first

src/components/layout/nav/components/nav-mobile.tsx
```diff
@@ -0,0 +1,106 @@
...
+                <LogoWrapper to="/" aria-label="Home">
+                    <img src={LogoOnly} alt="deriv logo" width={115} />
+                    <LogoDescription ai="center">
+                        <Line />
+                        <img src={GetTrading} alt="get trading" />
+                    </LogoDescription>
+                </LogoWrapper>
+
+                <LeftSection>
+                    {!hide_language_switcher && <LanguageSwitcher short_name="true" is_high_nav />}
```
yashim-deriv, line 78:
```suggestion
                    {!hide_language_switcher && <LanguageSwitcher short_name is_high_nav />}
```
Please refactor this as well even though it is slightly out of scope 🙏🏼  Using the string `"true"` for a boolean check feels super off. Also if you plan to do it refactor it to a recognisable boolean name: `is_short_name`

src/components/layout/nav/components/right-section.tsx
```diff
@@ -0,0 +1,99 @@
...
+    if (is_logged_in) {
+        return (
+            <Wrapper>
+                <Language hide_component={hide_language_switcher} />
+                <StyledButton onClick={handleGetTrading} primary>
+                    {localize('Get Trading')}
+                </StyledButton>
+            </Wrapper>
+        )
+    } else {
```
yashim-deriv, line 70:
```suggestion
    } else {
```
We do not need this `else` as there is already a `return` statement at the previous block and the following block would never be executed if the previous condition is met.

src/components/layout/nav/nav-academy.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const [prev_scroll_position, setPrevScrollPosition] = useState(0)
+    const [visible, setVisible] = useState(true)
+    const [is_mounted] = usePageLoaded()
+    const [is_logged_in] = useAuth()
+
+    const scrollHandler = () => {
+        const current_scroll_position = window.scrollY
+        setPrevScrollPosition(current_scroll_position)
+        setVisible(current_scroll_position > 72)
+    }
```
yashim-deriv, line 18:
```suggestion
    const scrollHandler = () => {
        const current_scroll_position = window.scrollY
        setPrevScrollPosition(current_scroll_position)
        setVisible(current_scroll_position > 72)
    }
```
Please freeze this function reference in a `useCallback`. Since it involves event subscription we need to ensure the removal of eventListener is done on the same function as it was subscribed.

src/components/layout/nav/nav.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const Nav = ({ is_ppc, is_ppc_redirect, hide_language_switcher, hide_signup_login }: NavProps) => {
+    const [is_logged_in] = useAuth()
+    const navProps = {
+        is_ppc,
+        is_logged_in,
+        is_ppc_redirect,
+        hide_language_switcher,
+        hide_signup_login,
+    }
```
yashim-deriv, line 22:
```suggestion
```

yashim-deriv (changes requested):
Few comments 🙏🏼

src/components/hooks/use-auth.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+import { isLoggedIn } from 'common/utility'
+
+const useAuth = () => {
```
yashim-deriv, line 4:
```suggestion
const useAuth = () => {
```
`useAuth` sounds misleading 🤔 we should name it something understandable like `useAuthCheck`

src/components/layout/nav/components/right-section.tsx
niloo-fs: So if the condition is false we don't have anything to return

niloo-fs: We have 2 separate components with 2 different conditions, if the first one return true we should render component A else component B

sean-binary (approved):

yashim-deriv (dismissed):
LGTM

sean-binary (approved):

PR #2662, opened 2022-02-17
refactor: add deriv be to domain list

sean-binary:
Changes:

-   add deriv be to domain list for redirection links

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -3,40 +3,52 @@ const isBrowser = () => typeof window !== 'undefined'
 const deriv_com_url = 'deriv.com'
 const deriv_me_url = 'deriv.me'
 const deriv_be_url = 'deriv.be'
+const staging_deriv_be_url = 'staging.deriv.be'
 
 export const deriv_com_app_id = 16929
 export const deriv_me_app_id = 1411
 export const deriv_be_app_id = 30767
 
-const supported_domains = [deriv_com_url, deriv_me_url, deriv_be_url]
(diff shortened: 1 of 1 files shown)


src/common/constants.ts
```diff
@@ -3,40 +3,52 @@ const isBrowser = () => typeof window !== 'undefined'
...
+                return deriv_me_url
+            case deriv_be_url:
+                return deriv_be_url
+            case staging_deriv_be_url:
+                return deriv_be_url
+            default:
+                return deriv_com_url
+        }
+    }
+}
```
prince-deriv, line 27: Can we have an object with a pair of domain name and it's corresponding url, so instead of using `switch` we can just get the keys of object and if the hostname exist we just use the value of it otherwise return the `deriv_com_url`

```
 domain_urls = {
   [deriv_com_url] : deriv_com_url,
   [deriv_me_url]: deriv_me_url
   // and so on 
   }
   
return Object.keys(domain_urls).includes(window.location.hostname)) ? domain_urls[window.location.hostname] : deriv_com_url

prince-deriv (approved):

njazuli-deriv (approved):

PR #2722, opened 2022-03-06
mehrzad / add dbot to multiplier

mehrzad-fs:


src/pages/trade-types/multiplier/_cfds.tsx
```diff
@@ -76,7 +76,7 @@ const CFDs = () => {
                     )}
                 </Text>
                 <AvailablePlatformsWrapper mb="40px">
-                    <AvailablePlatforms m_top="16px" dtrader />
+                    <AvailablePlatforms m_top="16px" dtrader dbot />
```
sean-binary, line 79: We'll need to add the check for `is_row` (Rest of world, AKA non-uk/eu) before displaying dbot as we don't show it for uk/eu.

mehrzad-fs: > We'll need to add the check for `is_row` (Rest of world, AKA non-uk/eu) before displaying dbot as we don't show it for uk/eu.

Thanks Sean 🙏 
Account type is not mentioned on the card description let me check it with PM

sean-binary (approved):
LGTM 👍

PR #2829, opened 2022-03-24
Revert "Proxy pass facebook pixel"

sean-binary:
Reverts binary-com/deriv-com#2821


yashim-deriv (approved):
LSTM (Looks sad to me)

PR #2834, opened 2022-03-25
NikitaK / Task - {update PA page design}

NikitaK-deriv:
Changes:

-   update PA page design and content

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [x] Translation to crowdin
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


alamin-deriv (approved):

src/components/layout/footer/main-links.js
```diff
@@ -152,23 +152,13 @@ const MainLinksSection = ({ is_ppc, is_ppc_redirect }) => {
...
-                            <Link
-                                to=""
-                                type="api"
-                                target="_blank"
-                                external="true"
-                                rel="noopener noreferrer"
-                            >
-                                {localize('API')}
-                            </Link>
+                            <Link to="https://api.deriv.com/">{localize('API')}</Link>
```
vijayasree-deriv, line 161: Won't hardcoding the API URL always redirect to live version? this will be an issue for example in the staging version, it will redirect to the live version instead of the staging version.

sean-binary: This is an external site so I don't see an issue here. But what we can optimise for is the domain itself like `.me` and `.be` so we should instead use `deriv_api_url` here that will handle that.

PR #2855, opened 2022-04-01
NikitaK/  Social network icons for Non -eu clients are shown above footer in mobile

NikitaK-deriv:
Changes:

-   fix: plecement of non-eu share iconsin footer

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/footer.js
+++ b/src/components/layout/footer.js
@@ -24,18 +24,18 @@ const Footer = ({ type, is_ppc, is_ppc_redirect, academy, no_footer_links }) => 
                         <MainLinksSection is_ppc={is_ppc} is_ppc_redirect={is_ppc_redirect} />
                     )}
                     <DisclaimerSection is_academy={academy} />
-                    <Mobile>
-                        <UKEU>
+                    <UKEU>
+                        <Mobile>
                             <BottomSocialSection type={type} />
-                        </UKEU>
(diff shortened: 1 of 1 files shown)


prince-deriv (dismissed):

src/components/layout/footer.js
```diff
@@ -25,15 +25,24 @@ const Footer = ({ type, is_ppc, is_ppc_redirect, academy, no_footer_links }) =>
                     )}
                     <DisclaimerSection is_academy={academy} />
                     <Mobile>
-                        <BottomSocialSection type={type} />
-                    </Mobile>
-                    <Desktop>
-                        <NonEU>
+                        <UKEU>
```
sean-binary, line 28: It would be better if `UKEU` can go above `Mobile` so that as long as it isn't UKEU, it won't render.

sean-binary, line 28: It would be better if `UKEU` can go above `Mobile` so that as long as it isn't UKEU, it won't render.

prince-deriv (approved):

PR #2865, opened 2022-04-04
nginx image validation

wrasheed-deriv:
Changes:

-   Nginx image will be validated after build. If the validation won't get successful it will abort the build before image push.

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -94,6 +94,16 @@ jobs:
             - name: Building docker image 🐳
               run: docker build -t ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:latest -t ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:$GIT_TAG_NAME .
 
+            - name: Verify nginx image
+              run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/production.yml
```diff
@@ -94,6 +94,16 @@ jobs:
...
 
+            - name: Verify nginx image
+              run: |
+                  nginx_test=`docker run --rm -it --entrypoint bash ${{ secrets.DOCKHUB_ORGANISATION }}/deriv-com:latest-staging -c "nginx -t" | grep -ow "test is successful"`
+                  if [[ $nginx_test == "test is successful" ]];then
+                    echo "Image validation successful"
+                    else
+                    echo "Image validation unsuccessful"
+                    exit 1
+                  fi
```
shudarshon-binary, line 105: we can use `set -e` to enforce error on exit. since our image `regentmarkets/deriv-com` doesn't have `bash` shell so we can avoid that. testing can be achieved even without introducing if-else block. even using `grep` is not needed

wrasheed-deriv: Thanks for the suggestions. Changes have been pushed.

shudarshon-binary (approved):

PR #2964, opened 2022-04-21
Ashraf lazyload intersection api

ashraf-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/hooks/use-handle-lazy-load.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ReactNode, useEffect, useState } from 'react'
+type templateProps = {
+    lazytemplates?: ReactNode
+    targetId?: string
+    options?: object
+}
+const UseHandleLazyLoad = ({ lazytemplates, targetId, options }: templateProps) => {
```
ali-hosseini-fs, line 7: since this is a component and not a custom hook please rename it to something like `IntersectionApiLazyloader`
convention of `useSomething` is just for custom hooks

package.json
```diff
@@ -79,6 +79,7 @@
         "eslint-webpack-plugin": "^3.1.1",
         "gatsby-plugin-eslint": "^4.0.2",
         "gatsby-plugin-styled-components": "^5.9.0",
+        "gatsby-plugin-webpack-bundle-analyser-v2": "^1.1.27",
```
ali-hosseini-fs, line 82: please don't commit bundle analyzer

gatsby-config.js
```diff
@@ -227,5 +227,11 @@ module.exports = {
             },
         },
         'gatsby-plugin-use-query-params',
+        {
```
ali-hosseini-fs, line 230: you can remove it
we don't need it on prod

src/components/hooks/use-handle-lazy-load.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ReactNode, useEffect, useState } from 'react'
+type templateProps = {
+    lazytemplates?: ReactNode
```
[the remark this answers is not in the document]
ali-hosseini-fs: and also it should not be optional, so please remove `?` wherever you can

src/components/hooks/use-handle-lazy-load.tsx
```diff
@@ -0,0 +1,40 @@
...
+            setIsVisible(entry.isIntersecting)
+            observer.disconnect()
+        }
+    }
+    useEffect(() => {
+        createObserver()
+        if (window.innerHeight >= 895) {
+            setLargeScreen(true)
+        }
+    }, [])
```
ali-hosseini-fs, line 29: this empty dependency array throws error on the console. 
you need to add createObserver to it
`}, [createObserver])`

src/components/hooks/use-handle-lazy-load.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const createObserver = () => {
+        const target = document.querySelector(targetId)
+        const observer = new IntersectionObserver(handleIntersect, options)
+        observer.observe(target)
+    }
+
+    return (isVisible || is_large_screen) && <>{lazytemplates}</>
```
ali-hosseini-fs, line 37: you can write this part better since if two first conditions fails it will return false there
so we need to add the loading if the condition were not applied
maybe something like this

```
if (isVisible || is_large_screen) return lazyTemplate
   return <LoadingComponent/>
```

src/pages/index.tsx
```diff
@@ -1,17 +1,67 @@
...
     /* redirect livechat for en to open live chat popup */
     useOpenLiveChat()
-
+    const target = '#market-fold'
+    const options = {
+        root: null,
+        rootMargin: '-50px',
+        threshold: 0.1,
+    }
+    const templates: ReactNode = (
```
ali-hosseini-fs, line 24: please choose a better name for this
Maybe `LazyComponents` ?

src/components/hooks/use-handle-lazy-load.tsx
ashraf-deriv: I guess it should be fine ,since it is supposed to execute only once after component render.
If is there any VP height changes, interceptor will handle that.

PR #3012, opened 2022-05-06
Aswathy/fix: removal of affiliate token from cookies

aswathy-deriv:
Changes:

-   Removed the affiliate token from cookies during signup without affiliate token

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/signup.js
+++ b/src/components/custom/signup.js
@@ -91,9 +91,9 @@ const Signup = (props) => {
 
         if (!token) {
             Cookies.remove('affiliate_tracking')
-            cookies_value.utm_campaign = 'null' //passing null as the cookies takes the affiliates token value when signed up without affiliate token
-            cookies_value.utm_medium = 'null'
-            cookies_value.utm_source = 'null'
+            delete cookies_value.utm_campaign
+            delete cookies_value.utm_medium
+            cookies_value.utm_source = 'null' //passing null as the cookies takes the affiliates token value when signed up without affiliate token
         }
 
         return {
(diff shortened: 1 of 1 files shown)


src/components/custom/signup.js
```diff
@@ -89,6 +89,14 @@ const Signup = (props) => {
         const cookies = getCookiesFields()
         const cookies_objects = getCookiesObject(cookies)
         const cookies_value = getDataObjFromCookies(cookies_objects, cookies)
+        const token = queryParams.get('t')
+
+        if (token == null) {
```
prince-deriv, line 94:
```suggestion
        if (!token) {
```

src/components/custom/signup.js
```diff
@@ -89,6 +89,14 @@ const Signup = (props) => {
...
+            Cookies.remove('affiliate_tracking')
+            cookies_value.utm_campaign = 'null' //passing null as the cookies takes the affiliates token value when signed up without affiliate token
+            cookies_value.utm_medium = 'null'
+            cookies_value.utm_source = 'null'
```
prince-deriv, line 98: We don't want the  `utm_campaign` and `utm_medium` show up in the API request. We can assign `null` instead of 'null' to eliminate them

prince-deriv (dismissed):

src/components/custom/signup.js
aswathy-deriv: validation fails if we pass null.so that's the reason i passed as string null

prince-deriv (approved):

prince-deriv (approved):

PR #3185, opened 2022-06-09
Mohsen / reorganized deps and dev deps

michio-rework:
Changes:
- removed react-is and webpack ( we don't need it directly in our deps I guess ) since we are not using them. 
- moved eslint-import-resolver-node, cli-select, commander

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [x] Dependency update
-   [ ] Documentation update
-   [ ] Release


package.json
```diff
@@ -47,15 +44,16 @@
         "react-dom": "^17.0.2",
         "react-helmet": "^6.1.0",
         "react-i18next": "^11.15.3",
-        "react-is": "^17.0.2",
```
ali-hosseini-fs, line 50: dont we need this one anymore?

package.json
```diff
@@ -47,15 +44,16 @@
...
         "react-paginate": "^8.1.0",
         "react-perfect-scrollbar": "^1.5.8",
         "react-ticker": "^1.3.1",
         "use-query-params": "^1.2.3",
-        "web-push-notifications": "^3.24.0",
-        "webpack": "^5.65.0"
```
ali-hosseini-fs, line 56: we need to check that removing this does not affect anything!

package.json
michio-rework: I searched the whole project, it's not used anywhere. should we keep it?

package.json
michio-rework: I checked around as far as I could, everything was fine, how about we add it in devDeps instead? 
definitely we don't need webpack in deps.

ali-hosseini-fs: we can pass it to qa to test it completely
if they dont find any issues then we can proceed with it

package.json
ali-hosseini-fs: no, its ok,
lets remove it

package.json
michio-rework: cool, will request their help on this

ali-hosseini-fs (approved):
LGTM

habib-deriv (dismissed):

habib-deriv (approved):
LGTM

PR #3187, opened 2022-06-09
Mohsen / added webpack bundle analyzer

michio-rework:
Changes:

- added webpack bundle analyzer gatsby plugin
- added analyze-bundle to generate report in webpack-bundle-report
- this won't affect the build process and only generates the report on running the script.
- added this new folder to .gitignore.

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [x] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,6 @@ yarn-error.log
 
 #CNAME
 static/CNAME
+
+#Webpack
+webpack-bundle-report
--- a/.gitignore
+++ b/.gitignore
@@ -82,7 +82,3 @@ yarn-error.log
 
 #CNAME
 static/CNAME
-
-
-# webpack bundle size reports
-/webpack-bundle-report
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,7 @@ yarn-error.log
 
 #CNAME
 static/CNAME
+
+
+# webpack bundle size reports
+/webpack-bundle-report
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     ],
     "license": "Apache License 2.0",
     "scripts": {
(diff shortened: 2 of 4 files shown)


habib-deriv (approved):
LGTM

ali-hosseini-fs (commented):
i think this runs on each build
so maybe its better to add a script for generating the report in package.json

michio-rework:
> i think this runs on each build so maybe its better to add a script for generating the report in package.json

thanks for the suggestion, added the npm script for generating report.

.gitignore
```diff
@@ -81,4 +81,4 @@ yarn-error.log
 .idea/
 
 #CNAME
-static/CNAME
+static/CNAME
```
ali-hosseini-fs, line 84: there is an error for EOL
please add a new line at the end of this

package.json
```diff
@@ -103,6 +104,7 @@
     ],
     "license": "Apache License 2.0",
     "scripts": {
+        "analyze-bundle": "GENERATE_JSON_STATS=true gatsby build && npx webpack-bundle-analyzer ./public/stats.json",
```
ali-hosseini-fs, line 107: where does this save generated reports?
does it generate HTML file like before?

.gitignore
michio-rework: thanks, sure will do.

package.json
michio-rework: it does not generate the file at the moment, it just opens a server with the actual report. will change it to generate the file as well.

ali-hosseini-fs (approved):

mohammad-rework (approved):

PR #3261, opened 2022-06-23
Mohsen / fixed menu drop down arrows

michio-rework:
Changes:
-  fixed arrow in menu drop down navigation bar 

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/custom/platforms-dropdown.js
+++ b/src/components/custom/platforms-dropdown.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import styled, { keyframes } from 'styled-components'
 import {
@@ -22,7 +22,7 @@ const FadeInDown = keyframes`
 const NavDropdown = styled.div`
     width: auto;
     max-width: 1200px;
-    left: ${(props) => (props.offset && !props.is_trade ? props.offset + 'px !important' : 'none')};
+    left: ${(props) => (props.offset ? props.offset + 'px !important' : 0)};
     position: absolute;
     padding: 2.2rem 0.8rem;
     z-index: -1;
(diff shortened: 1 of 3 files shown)


mohammad-rework (approved):

src/components/custom/platforms-dropdown.js
```diff
@@ -77,47 +67,36 @@ const PlatformsDropdown = ({
...
     const is_trade = active_dropdown === 'trade'
-    const current_offset = current_ref.offsetWidth / 2 - 15
-    const setTradeArrowOffset = (dropdownOffset) =>
-        current_ref.offsetLeft - dropdownOffset + current_offset
-    const [left_offset, setLeftOffset] = useState(current_ref.offsetLeft)
-    const [left_arrow_offset, setLeftArrowOffset] = useState()
+
+    const [left_offset, setLeftOffset] = useState(() => {
+        if (is_trade) {
+            return current_ref?.getBoundingClientRect()?.x / 2
```
mitra-deriv, line 73: Hi @michio-rework, just wanna ask what this part of the code is doing to understand it better ? 🙏

michio-rework: Hi @mitra-deriv , please check [this](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) out.

I'm using `X` coordinates of the clicked `li` element as `left` property of dropdown ( the absolute div which is being opened ). 

But for the first `li` ( Trade ) we have show it differently since based on ROW and etc we have different items in dropdown and eventually the size of it will be bigger. in smaller screen if we use the `X` coordinate of the `li` when region is ROW it goes our of screen sometimes. so we have to use half of `X` coordinate of clicked `li` to make sure the content is matched with the desing. 

If you have more questions please dm me and we can discuss this furthur.

PR #3336, opened 2022-07-18
MohammadH/ Wizard Component

mohammad-rework:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/form/wizard/footer.tsx
```diff
@@ -0,0 +1,44 @@
...
+                tertiary="true"
+                disabled={step <= 1}
+                onClick={() => step > 1 && setStep(step - 1)}
+            >
+                {localize('Previous')}
+            </Button>
+            <Button
+                secondary="true"
+                disabled={disabled}
+                onClick={() => step < 5 && setStep(step + 1)}
```
michio-rework, line 36: can we create a function for `onClick`s, this function can have "previous" and "next" as enum or something like that?

src/components/custom/_card.tsx
```diff
@@ -0,0 +1,72 @@
...
+const SelectedIcon = styled.img`
+    position: absolute;
+    top: 16px;
+    right: 16px;
+    width: 16px;
+    height: 16px;
+`
+
+const Card = ({ icon, title, description, selected }: CardProps) => {
+    const [is_selected, setIsSelected] = React.useState(selected || false)
```
michio-rework, line 57: using props as intialValue of state is anit-pattern to be honest, since the `selected` prop can change weird things can happen, for more information and why this should be avoided please check this out : https://vhudyma-blog.eu/react-antipatterns-props-in-initial-state/
we can talk about it more if you need.

src/components/custom/_card.tsx
```diff
@@ -54,9 +54,15 @@ const SelectedIcon = styled.img`
 `
 
 const Card = ({ icon, title, description, selected }: CardProps) => {
-    const [is_selected, setIsSelected] = React.useState(selected || false)
+    const [is_selected, setIsSelected] = React.useState(false)
     const clickHandler = () => setIsSelected(true)
 
+    React.useEffect(() => {
+        if (selected) {
```
michio-rework, line 61: do we need this if condition here? 
initially the state will be `false` so there will be no changes.

src/components/form/wizard/footer.tsx
```diff
@@ -18,22 +18,37 @@ const StyledFooter = styled.div`
...
+const enum ButtonType {
+    Previous = 'PREVIOUS',
+    Next = 'NEXT',
+}
+
 const Footer = ({ disabled }: { disabled: boolean }) => {
-    const { step, setStep, max_step } = React.useContext(StepContext)
+    const { step, setStep, max_step, setEnableNext } = React.useContext(StepContext)
+
+    const buttonHandler = (button_type: ButtonType): void => {
```
michio-rework, line 29: can you please use `useCallback` here? 
since you are getting some values from context, it may re-generate the function again on every render.

michio-rework (approved):

PR #3409, opened 2022-07-28
NikitK/ Convert containers folder to ts 

NikitK-deriv:
Changes:

-   convert containers folder to ts

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


mohammad-rework (dismissed):

src/components/containers/show.tsx
```diff
@@ -0,0 +1,45 @@
...
+    min_width?: string & number
+}
+
+export const Desktop = ({ children, max_width, ...props }: ResponseType) => (
+    <Media greaterThanOrEqual={max_width || 'tabletL'} {...props}>
+        {children}
+    </Media>
+)
+
+// We can't type it properly due fresnel
```
aswathy-deriv, line 17: is this comment needed?

src/components/containers/show.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useContext, ReactNode } from 'react'
+import { Media } from 'themes/media'
+import { DerivStore } from 'store'
+
+type ResponseType = {
+    children?: ReactNode
+    max_width?: string & number
```
aswathy-deriv, line 7: we can use max_width:string | number right

src/components/containers/show.tsx
```diff
@@ -0,0 +1,45 @@
...
+    min_width?: string & number
```
aswathy-deriv, line 8: here also.instead of and operator use or

src/components/containers/show.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const { is_eu_country } = useContext(DerivStore)
+
+    if (is_eu_country) return <>{children}</>
+    else return null
+}
+
+export const NonEU = ({ children }: DomainType) => {
+    const { is_eu_country } = useContext(DerivStore)
+
+    if (is_eu_country === false) return <>{children}</>
```
maryia-binary, line 36: small suggestion:
```suggestion
    if (!is_eu_country) return <>{children}</>
```
is it gonna work if you `return children` without wrapping it in a fragment <></> ?  😊

src/components/containers/show.tsx
NikitK-deriv: Unfortunately no.
We use dependency fresnel which should include `greaterThanOrEqual` which include string and number type in node_modules. I'm not yet realized how to change it
Screenshot 2022-07-28 at 15 00 41

src/components/containers/show.tsx
NikitK-deriv: Seems we should explain why we decide to use any here

src/components/containers/show.tsx
NikitK-deriv: Resolved as NonEU = Eu
Because components are the same 💩

mohammad-rework (approved):

PR #3437, opened 2022-08-02
Mitra/Fix sequence of locations

mitra-deriv:
Changes:

-   ...

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/elements/default-tab.js
```diff
@@ -121,12 +121,13 @@ const Tabs = ({
     mobile_font_size,
     mobile_tab_button_underline_length,
     has_no_query,
+    starting_index = 0,
```
iman-fs, line 124: just a suggestion:
instead of passing the index, you can use the value of the tab from tab_list. because this way, if they decide to change the order of tab items, you don't need to change the logic here again.

iman-fs (commented):
.

iman-fs (approved):

michio-rework (approved):

PR #3517, opened 2022-08-22
NikitK/ Convert typography and link files to ts

NikitK-deriv:
Changes:

-   convert typography and link to typescript

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/elements/typography.ts
+++ b/src/components/elements/typography.ts
@@ -1,9 +1,22 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
-import PropTypes from 'prop-types'
 import { generateResponsiveStyles } from '../containers/box'
-import { Margins, Paddings } from 'themes/function'
+import { Margins, MarginsType, Paddings, PaddingsType } from 'themes/function'
 import device from 'themes/device'
+
+type BaseElementProps = {
+    align?: string
+    padding?: string
+    color?: string
+    lh?: string
+    max_width?: string
+    min_width?: string
+    min_height?: string
+    max_height?: string
+    mobile_max_width?: string
+} & MarginsType &
+    PaddingsType &
+    React.CSSProperties
 
 const baseStyles = ({
     m,
@@ -24,7 +37,7 @@ const baseStyles = ({
     max_height,
     width,
     height,
-}) => css`
+}: BaseElementProps) => css`
     min-width: ${min_width};
     max-width: ${max_width};
     min-height: ${min_height};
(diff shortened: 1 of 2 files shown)


hamid-deriv (dismissed):

prince-deriv (dismissed):

src/components/elements/typography.ts
```diff
@@ -85,7 +100,13 @@ export const Text = styled.p`
 /////////////////// TEXT IS DEPRECATED. PLEASE USE HEADER. ///////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-export const Header = styled(({ as = 'h2', children, ...props }) =>
+type HeaderProps = {
+    as?: string
```
michio-rework, line 104:
```suggestion
    as?: React.ElementType;
```
what do you think about this approach @NikitK-deriv ? 
this is just a suggestion, but if you add it we will get better type inference :pray:

src/components/elements/typography.ts
```diff
@@ -85,7 +100,13 @@ export const Text = styled.p`
...
+    children?: React.ReactNode
+    props?: TextProps
```
michio-rework, line 106: do we have to add `props` here? 
since you are adding everything from `TextProps` to the type, I think we omit this. 
If I'm missing something, please let me know, thanks thanks :pray:

NikitK-deriv: We might need to work on our storybook later. It's answer for our team
Thanks for good question, I thought the same 🙇

src/components/elements/typography.ts
NikitK-deriv: Unfortunately we can't, because pass a string only here. To make seo optimization better. And sharing what exatcly I get by this way. Also I tried `React.ElementType | string` but no luck with that
Screenshot 2022-09-06 at 12 04 57

src/components/elements/typography.ts
```diff
@@ -85,7 +100,13 @@ export const Text = styled.p`
...
+} & TextProps
+
+export const Header = styled(({ as = 'h2', children, ...props }: HeaderProps) =>
```
mohammad-rework, line 109: We are getting typescript errors in many files where we used `<Header>` component.
Try to fix those errors.

NikitK-deriv: Already resolved. Just forgot to add generic for styled component after props. Let's check again 🙏

src/components/elements/typography.ts
NikitK-deriv: Right. Also decided to merge `TextProps` to `BaseElementProps`
Thanks 🙇

src/components/elements/typography.ts
mohammad-rework: I still see the errors.
check these files: 
`src/pages/endpoint/index.tsx`
`src/pages/home/_hero.tsx`

PR #3604, opened 2022-09-06
feat: added stylis plugin for logical properties

michio-rework:
Changes:
- added stylis plugin in `plugin` folder 
- with this approach we don't have to change our styles and the plugin will handle the transformation to `logical css properties` on it's own. 
- this is the basic config setup and more stuff must be handled in JS world since there are some logic behind them.  
- the browser support is good enough IMHO you can check it [here](https://caniuse.com/css-logical-props)
- As of now, styled-componentv5.x uses stylis v3.x, styled-component v6.x uses stylis v4.x and the new version has better plugin features. they reworked the whole plugin api and interface in stylis v4.x
 
## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/elements/off-canvas-menu.js
```diff
@@ -57,7 +58,15 @@ const OffCanvasMenu = styled.section`
     transition: transform 0.4s;
     box-shadow: 0 16px 20px 0 rgba(0, 0, 0, 0.1);
     left: -254px;
-    ${({ is_canvas_menu_open }) => is_canvas_menu_open && 'transform: translateX(254px)'};
+    ${({ is_canvas_menu_open, is_rtl }) => {
```
habib-deriv, line 61: Can we make a function for this for these translate issues?

michio-rework: added `withLangDirection` function, thanks thanks :pray:

src/components/containers/seo.js
```diff
@@ -92,6 +93,9 @@ const SEO = ({ description, meta, title, no_index, has_organization_schema, meta
             htmlAttributes={{
                 lang: formatted_lang,
             }}
+            bodyAttributes={{
+                dir: i18next.dir(i18next.language),
```
prince-deriv, line 97: Not sure how often we need to get the direction of the language but can we have a quick helper function of this?

michio-rework: you are absolutely right @prince-deriv , in fact I have the function, didn't use it. will use it here as well, thanks :pray:

src/components/localization/config.js
```diff
@@ -27,4 +27,20 @@ i18n.languages = Object.keys(language_config)
 
 export const localize = (string, values) => t(crc32(string), { defaultValue: string, ...values })
 
+/**
+ * @description if you need the direction of the current languge please use this fucntion ( without react life-cycles), if you need react life-cycles please use useLangDirection hook
+ * @returns {"ltr" | "rtl"} the current language direction
+ */
+export const get_lang_direction = () => {
+    return i18n.dir(i18n.language)
```
farrah-deriv, line 35: do we need to pass the language here since we want to get the dir of the current language?

michio-rework: Thanks @farrah-deriv , I can remove this. cool :pray:

i18n-config.js
```diff
@@ -85,6 +85,11 @@ module.exports = {
         short_name: 'TR',
         affiliate_lang: 14,
     },
+    ar: {
+        path: 'ar',
+        display_name: 'Arabic',
```
yashim-deriv, line 90:
```suggestion
        display_name: 'Arabic',
```
`Arabic` string needs to be in its translated form. `عَرَبِيّ` or something like that

src/themes/function.ts
```diff
@@ -16,17 +17,30 @@ export type PaddingsType = {
...
+    ltr_styles: css``,
+    rtl_styles: css``,
+}
+
+export const withLangDirection = (styles = default_styels) => {
+    const { rtl_styles, ltr_styles } = styles
+    if (is_rtl()) {
+        return rtl_styles
+    }
+    return ltr_styles
```
yashim-deriv, line 45:
```suggestion
    return is_rtl() ? rtl_styles : ltr_styles;
```

src/themes/plugin/margin.ts
```diff
@@ -0,0 +1,18 @@
+export const handleMargin = (margin: string) => {
+    if (margin.includes('calc') || margin.includes('var')) {
+        return `margin: ${margin};`
+    }
+    const margins = margin?.split?.(' ')
+
+    switch (margins?.length) {
```
yashim-deriv, line 7:
```suggestion

    const margins = margin?.split?.(' ')
    switch (margins?.length) {
```

Space consistency

src/themes/plugin/padding.ts
```diff
@@ -0,0 +1,17 @@
+export const handlePadding = (padding: string) => {
+    if (padding.includes('calc') || padding.includes('var')) {
+        return `padding: ${padding};`
+    }
+    const paddings = padding?.split?.(' ')
+    switch (paddings?.length) {
```
yashim-deriv, line 6:
```suggestion
    }
    
    const paddings = padding?.split?.(' ')
    switch (paddings?.length) {
```

src/themes/plugin/index.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @param  {number} line
+ * @param  {number} column
+ * @param  {number} length
+ * @return {(string|void)?}
+ */
+export const plugin = (context, content) => {
+    switch (context) {
+        case 1:
+            return handle(content)
+
```
yashim-deriv, line 87:
```suggestion
```

src/themes/plugin/index.ts
```diff
@@ -0,0 +1,98 @@
+import { handleMargin } from './margin'
+import { handlePadding } from './padding'
+
+const handle = (content: string) => {
```
yashim-deriv, line 4:
```suggestion
const transformCSSProperties = (content: string) => {
```
Probably need a better name to explain what it is doing better

src/themes/global-style.ts
```diff
@@ -35,6 +35,12 @@ const GlobalStyle = createGlobalStyle`
         padding: 0;
         background: var(--color-black);
     }
+    *[dir='rtl'] {
```
yashim-deriv, line 38:
```suggestion
    *[langdirection='rtl'] {
```
Can we rename it to langdir or langdirection? It will be much clearer on what it does

yashim-deriv (commented):
Some nits and comments. Overall, nice job! There might be a way to do it during build time by extending styled components but that would be overkill and take too long 😅

i18n-config.js
michio-rework: thanks, it should be `العربية`, fixed it. :pray:

src/themes/global-style.ts
michio-rework: unfortunately we can't since we are setting the `dir` html attribute with `helmet`, but I'll provide a comment on top it and explain what it does.

habib-deriv (dismissed):
LGTM

src/components/elements/image-with-direction.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import styled from 'styled-components'
+import { useIsRtl } from 'components/hooks/use-isrtl'
+
+export type TImageWithDirection = {
+    is_rtl: boolean
```
mohammad-rework, line 6: beside `is_rtl` can we add necessary types for image such as `alt` `width` `height` to be sure everyone is using these props as well while using `ImageWithDirection` component? If typescript already throws an error when we didn't pass those props so please skip this comment.

i18n-config.js
```diff
@@ -85,6 +85,11 @@ module.exports = {
         short_name: 'BN',
         affiliate_lang: 15,
     },
+    ar: {
+        path: 'ar',
+        display_name: 'العربية',
+        short_name: 'العربية',
```
habib-deriv, line 91: Can we change this short name to 'AR' probably?

prince-deriv:
Can we also fix the small code smells here?

i18n-config.js
michio-rework: done :+1:

src/components/elements/image-with-direction.tsx
michio-rework: I think it's not cool to force people to pass `width`, `alt`, the type checking for component props works. so it's already there.

michio-rework:
> 

I fixed the ones I could, the last one is about Using `Header` and `Text` is deprecated. 
@prince-deriv

habib-deriv (approved):
LGTM

yashim-deriv (approved):

PR #3658, opened 2022-09-20
Eduard / Attribute loading not allowed on element div (loading isn't an attribute, it's a props)

eduard-deriv:
Changes:

-   removed unused code in ImageWrapper (never changes `disable_transition` props value to `true`, always uses default `disable_transition` props value, `loading` value changes due to `[data-main-image]`. Both conditions must be met, but this behavior will never happen)

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ X ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/elements/query-image.tsx
+++ b/src/components/elements/query-image.tsx
@@ -1,26 +1,17 @@
-import React, { CSSProperties, MouseEventHandler, ReactElement } from 'react'
+import React, { CSSProperties, ReactElement } from 'react'
 import styled from 'styled-components'
 import { GatsbyImage, getImage } from 'gatsby-plugin-image'
 import type { ImageDataLike, IGatsbyImageData, GatsbyImageProps } from 'gatsby-plugin-image'
 
-type CustomOmit = Omit<
-    GatsbyImageProps,
-    'width' | 'height' | 'className' | 'style' | 'onMouseOver' | 'onMouseOut' | 'onClick'
->
-
 export type QueryImageProps = {
     alt: ReactElement | string
-    className?: string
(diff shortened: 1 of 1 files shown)


src/components/elements/query-image.tsx
```diff
@@ -58,11 +44,9 @@ const QueryImage = ({
     if (data) {
         return (
             <ImageWrapper
-                loading={loading}
```
yashim-deriv, line 61: The `loading` property is still used in multiple places. Some functionality will break if you remove this property. 

A few explanation for this property: It is an extension of the base type of `gatsby-plugin-image` https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/#shared-props

You can check this by searching for `loading="eager"`

src/components/elements/query-image.tsx
```diff
@@ -22,32 +21,19 @@ export type ImageWrapperProps = {
     width: string
     height: string
     className?: string
-    loading: 'eager' | 'lazy'
```
yashim-deriv, line 25: Can you remove `width`, `height`, `className` and other properties that are already declared by the `IGatsbyImageData`? Change this to an interface and extend 

```
export interface QueryImageProps extends IGatsbyImageData
```

yashim-deriv (changes requested):

yashim-deriv (approved):

niloo-fs (approved):
LGTM

PR #3682, opened 2022-09-29
Mitra/Pr 75267 poc for marking component as deprecated

mitra-deriv:
### Proposal: Marking components as deprecated 
---

Special thanks to @farzin-fs for the idea and guidance 💡 🌸 

---
## Problem 
The `<text>` component was deprecated inside our code by only a comment which many developers can overlook and use it inside the code still. 

`Note that we can use this Plugin for any of our deprecated functions, components, etc..`

## Solution
As we are using this component in many places refactoring the code may take time so I came with a POC to mark the component as deprecated so it shows a warning to the developers whenever they want to use components which are deprecated so they won’t use this in new places (or even help to refactor in those files).

We found this [Plugin](https://github.com/gund/eslint-plugin-deprecation) which we can add  ESLint rule that reports usage of deprecated code and give a warning. 

deprecate


src/components/elements/typography.ts
```diff
@@ -103,9 +103,7 @@ export const BaseElement = css<BaseElementProps>`
     }
 `
 
-//////////////////////////////////////////////////////////////////////////////
-/////////////////// TEXT IS DEPRECATED. PLEASE USE HEADER. ///////////////////
-//////////////////////////////////////////////////////////////////////////////
+/** @deprecated Use Header instead */
```
farzin-fs, line 106:
```suggestion
/** @deprecated Use `Header` component instead. */
```

farzin-fs (approved):

habib-deriv (dismissed):
LGTM

michio-rework (approved):

farzin-fs (approved):

prince-deriv (approved):

habib-deriv (approved):

yashim-deriv (approved):

PR #3784, opened 2022-10-31
NikitK/ Connection lost when idled for few minutes

NikitK-deriv:
Changes:

-    reload page if ws connection closed after user idle

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/common/websocket/api.ts
+++ b/src/common/websocket/api.ts
@@ -76,7 +76,10 @@ export default class DerivWS {
             let lastActionTaken = new Date().getTime()
             const checkLastAction = () => {
                 const now = new Date().getTime()
-                if (now - lastActionTaken > 500) {
+                if (now - lastActionTaken > 0) {
+                    window.removeEventListener('mousemove', checkLastAction)
+                    window.removeEventListener('touchstart', checkLastAction)
+                    window.removeEventListener('keydown', checkLastAction)
                     this.recursiveConnect()
                 }
                 lastActionTaken = now
--- a/src/common/websocket/api.ts
+++ b/src/common/websocket/api.ts
@@ -50,8 +50,13 @@ export default class DerivWS {
     }&brand=${brand_name.toLowerCase()}`
     connection: WebSocket
 
+    private ping = () => {
+        this.send({ ping: 1 })
+    }
+
     private openHandler = () => {
         if (this.connection.readyState === 1) {
+            setInterval(this.ping, 1000)
             this.connected.resolve()
         } else {
             setTimeout(this.openHandler, 50)
@@ -68,30 +73,12 @@ export default class DerivWS {
         }
     }
 
-    private recursiveConnect = () => {
-        this.connection = new WebSocket(this.socket_url)
-        this.connection.onopen = this.openHandler
-        this.connection.onmessage = this.messageHandler
-        this.connection.onclose = () => {
-            let lastActionTaken = new Date().getTime()
-            const checkLastAction = () => {
-                const now = new Date().getTime()
-                if (now - lastActionTaken > 0) {
-                    window.removeEventListener('mousemove', checkLastAction)
-                    window.removeEventListener('touchstart', checkLastAction)
-                    window.removeEventListener('keydown', checkLastAction)
-                    this.recursiveConnect()
-                }
-                lastActionTaken = now
-            }
-            window.addEventListener('mousemove', checkLastAction)
-            window.addEventListener('touchstart', checkLastAction)
-            window.addEventListener('keydown', checkLastAction)
-        }
-    }
-
     constructor() {
-        this.recursiveConnect()
+        const connection = new WebSocket(this.socket_url)
+        connection.onopen = this.openHandler
+        connection.onmessage = this.messageHandler
+
+        this.connection = connection
     }
 
     isConnectionClosed() {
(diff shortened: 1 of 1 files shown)


src/common/websocket/api.ts
```diff
@@ -54,6 +54,7 @@ export default class DerivWS {
         const connection = new WebSocket(this.socket_url)
         connection.onopen = this.openHandler
         connection.onmessage = this.messageHandler
+        connection.onclose = this.reloadPage
```
prince-deriv, line 57:
```suggestion
        connection.onclose = this.reloadHandler
```

balakrishna-binary, line 57: Reloading the page should not be preferred. Can we try to re-establish the WS connection?

NikitK-deriv: Yes @balakrishna-binary actually I'm trying to do it

NikitK-deriv: Done @balakrishna-binary please check one more time 🙏

NikitK-deriv: Made function recursive and depends on users activity. Please check again @prince-deriv 🙏

src/common/websocket/api.ts
```diff
@@ -98,4 +99,16 @@ export default class DerivWS {
...
+        const checkLastAction = () => {
+            const now = new Date().getTime()
+            if (now > lastActionTaken) window.location.reload()
+            else lastActionTaken = now
+        }
+        window.addEventListener('mousemove', checkLastAction)
+        window.addEventListener('touchstart', checkLastAction)
+        window.addEventListener('keydown', checkLastAction)
+    }
 }
```
jim-deriv, line 116: Hello @NikitK-deriv,

Is it possible to implement something like `send({ "ping": 1 }, callback)` that is triggered every `1.5s` or so, instead of reloading the page?

You could use something like `setInterval or setTimeout` for it.

Because I see in the API Playground, there is a `ping` endpoint I think that is meant for this purpose.

Please look into it. 🙂  Thank you

jim-deriv, line 116: Hello @NikitK-deriv,

Is it possible to implement something like `send({ "ping": 1 }, callback)` that is triggered every `1.5s` or so, instead of reloading the page?

You could use something like `setInterval or setTimeout` for it.

Because I see in the API Playground, there is a `ping` endpoint I think that is meant for this purpose.

Please look into it. 🙂  Thank you

src/common/websocket/api.ts
```diff
@@ -71,6 +58,44 @@ export default class DerivWS {
...
+
+    private recursiveConnect = () => {
+        this.connection = new WebSocket(this.socket_url)
+        this.connection.onopen = this.openHandler
+        this.connection.onmessage = this.messageHandler
+        this.connection.onclose = () => {
+            let lastActionTaken = new Date().getTime()
+            const checkLastAction = () => {
+                const now = new Date().getTime()
+                if (now - lastActionTaken > 500) {
```
maryia-binary, line 79: @NikitK-deriv please elaborate on why do we need to check for the last action taken and why do we need to use 500 here? if we want to reopen connection when it's closed, we can simply do this:
```suggestion
                this.connection.onclose = () => {
                    this.recursiveConnect()
                }
```
I believe we don't need any listeners here especially while we'll need to remove them too at some point, otherwise they will fire multiple number of times whenever each of the below events happens. If the connection is closed, it's obviously closed because the user was idle and didn't do any actions, right? Or are there any cases when the connection gets closed and we don't want to reopen it?
Also, I think the PR description is a bit out-dated, you are still saying that you're going to reload the page. Please check 🙂

NikitK-deriv: Thanks for note 🙇

src/common/websocket/api.ts
```diff
@@ -71,6 +58,47 @@ export default class DerivWS {
...
+        const response: DerivAPIResponse = JSON.parse(msg.data)
+        const reqId = response.req_id
+
+        if (reqId && reqId in this.pendingRequests) {
+            this.pendingRequests[reqId].resolve(response)
+            delete this.pendingRequests[reqId]
+        }
+    }
+
+    private recursiveConnect = () => {
```
michio-rework, line 71: I think this is not an optimal approach for keeping the connection alive, using the `ping/pong` method is less resource consuming, as we know the web socket connection will be close on 2-3 minutes mark, that's when the recursive connect is called. 
the web socket connection is based on `HTTP` and we `upgrade` the connection, so when you open a new connection with `recursiveConnect` there will be at least 2 `HTTP` request, one for initial connection and one for `upgrade` ( handshake ) there will be other requests for handshake validation and etc. please look up `web socket handshake` in google or check out [this article](https://dev.to/hgsgtk/learn-websocket-handshake-protocol-with-gorillawebsocket-server-10k9)

I needed to implement this in one of my tasks, it's not merged yet, but I did it like so: 
https://github.com/binary-com/deriv-com/pull/3679/files#:~:text=private%20ping%20%3D,%2C%201000)

we can talk more on this if you want :pray:

michio-rework (changes requested):

src/common/websocket/api.ts
NikitK-deriv: Decided to use your approach. Thanks for info and guidance 👍

michio-rework (approved):

PR #3834, opened 2022-11-15
feat: added the speaker page

aswathy-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/conference/speakers/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const Layout = styled.div`
+    background: var(--color-white);
+`
+
+const data = [
+    {
+        id: 'name_01',
+        image: Name1,
+        title: <Localize translate_text="Name" />,
```
michio-rework, line 19:
```suggestion
type TSpeaker = {
    id: string,
    image: string,
    title : TString,
    content: TString
}
const data : TSpeaker[] = [{
    id: "name_01",
    image: Name1,
    title: "Name",
    content: "Who is the speaker? This is where you can write a bio to detail your speaker’s credentials and the unique perspective they will give to the conference."
    }
}]
```
we can't have this kinda structure now, filtering content and applying changes will be easier, can you please update it, thanks thanks :pray: 
> Note: Please check the next suggestion as well.

src/pages/conference/speakers/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        content: (
+            <Localize translate_text="Who is the speaker? This is where you can write a bio to detail your speaker’s credentials and the unique perspective they will give to the conference." />
+        ),
+    },
+]
+
+const Speakers = () => {
+    return (
+        <Layout>
+            {data.map(({ image, title, content, id }) => {
```
michio-rework, line 53:
```suggestion
            {data.map(({ image, title, content, id }) => {
               return <ImageContentBlock title={localize(title)} image={image} content={localize(content)} key={id} /> 
            }}
```
you can go further and update the `ImageContentBlock` component, so it can get the `TString` types and use the `localize` or `Localize` inside of it. it's gonna be even better, thanks :pray:

michio-rework (changes requested):
the image with `name1.svg` should be updated I guess. can we use `.png` or `.jpeg` files here instead of svg files?
thanks :pray:

PR #3852, opened 2022-11-16
feat: added use media-query

michio-rework:
Changes:
- added new hook, `use-media-query`, matchMedia is much better than added `resize` event listeners, this is gonna replace `use-window-size` in most scenarios ( not all of them, we need hook for now ) , check [this](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) out 
- added `Visibility` component, which is connected to `device` object and `use-media-query`, there is an example usage there in the PR as well. 
- the main difference in performance of this hooks vs `use-window-size` is, we let the browser inform us on `media` changes, instead of subscribing to the window's `resize` event. so the component is imperatively connected to the window size.  
- we have to update our devices object ( media-queries ), which will happen in the future. 

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


mohammad-rework (approved):

habib-deriv (dismissed):

src/themes/device.ts
```diff
@@ -25,7 +25,7 @@ const device = Object.freeze({
     tabletM: `(min-width: ${size.mobileL}px) and (max-width: ${size.tabletL}px)`,
     tabletS: `(max-width: ${size.tabletS}px)`,
     tablet: `(max-width: ${size.tablet}px)`,
-    tabletL: `(max-width: ${size.tabletL}px)`,
+    tabletL: `(max-width: ${size.tabletL - 1}px)`,
```
prince-deriv, line 28: Won't this affect alot of UI from this size?

michio-rework: you're right, this has side-effects. but we're gonna introduce new object for the useMediaQuery. I'll revert this back to normal. thanks man @prince-deriv :pray:

src/components/hooks/use-breakpoints.tsx
```diff
@@ -0,0 +1,28 @@
+import useMediaQuery from './use-media-query'
+import { breakpoints } from 'themes/theme.breakpoints'
+
+export type TUseBreakpoints = {
```
farzin-deriv, line 4: @michio-rework I don’t think we need to export it, If we need it somewhere we can use `ReturnType<typeof useBreakpoints>`

src/components/hooks/use-breakpoints.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const is_mobile = useMediaQuery(breakpoints.xs)
+    const is_tablet = useMediaQuery(breakpoints.sm)
+    const is_lg_tablet = useMediaQuery(breakpoints.md)
+    const is_laptop = useMediaQuery(breakpoints.lg)
+    const is_desktop = useMediaQuery(breakpoints.xl)
+
+    return {
+        is_mobile,
+        is_tablet,
+        is_lg_tablet,
```
farzin-deriv, line 22: @michio-rework [LG](https://www.lg.com/) or [Large](https://dictionary.cambridge.org/dictionary/english/large)? 😁 Let's avoid abbreviations 🙏🏻

src/components/hooks/use-visibility.tsx
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react'
+import { TUseBreakpoints } from './use-breakpoints'
+import { DerivStore } from 'store'
+
+// HINT: We will add more values to the hook in the future,
+// so for the type handling for now I'll just alias the type
+export type TVisiblity = TUseBreakpoints
```
farzin-deriv, line 7: @michio-rework Same here, Can use `ReturnType` generic if needed.

src/components/hooks/use-visibility.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+const useVisibility = (): TVisiblity => {
+    const { breakpoints } = useContext(DerivStore)
+
+    return breakpoints
```
farzin-deriv, line 12: @michio-rework I got a little confused here, Maybe it’s because of the namings. I see you will be adding more stuff to this hook later but couldn’t we just use `useBreakpoints` here instead of `useContext`? It returns the same thing, Right 🤔

src/components/hooks/use-breakpoints.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const useBreakpoints = (): TUseBreakpoints => {
+    const is_mobile = useMediaQuery(breakpoints.xs)
+    const is_tablet = useMediaQuery(breakpoints.sm)
+    const is_lg_tablet = useMediaQuery(breakpoints.md)
+    const is_laptop = useMediaQuery(breakpoints.lg)
+    const is_desktop = useMediaQuery(breakpoints.xl)
+
+    return {
```
farzin-deriv, line 19: @michio-rework Maybe it will be too much but we could also return values like `is_mobile_and_tablet` and etc here to avoid having repetitive `&&` or `||` conditions in the code.

src/pages/trade-types/cfds/_start-trading.tsx
```diff
@@ -4,29 +4,29 @@ import { SmallContainer, StartTradingBg } from '../components/_style'
...
 import { Header } from 'components/elements'
-import { Desktop, Mobile } from 'components/containers'
 import { useCountryRule } from 'components/hooks/use-country-rule'
+import useVisibility from 'components/hooks/use-visibility'
 
 const StartTrading = () => {
     const { is_eu } = useCountryRule()
+    const viz = useVisibility()
+    const { is_mobile, is_tablet } = viz
     return (
```
farzin-deriv, line 14:
```suggestion
    const visibility = useVisibility()
    const { is_mobile, is_tablet } = visibility
    
    return (
```

src/components/hooks/use-visibility.tsx
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react'
+import { TUseBreakpoints } from './use-breakpoints'
+import { DerivStore } from 'store'
+
+// HINT: We will add more values to the hook in the future,
+// so for the type handling for now I'll just alias the type
+export type TVisiblity = TUseBreakpoints
+
+const useVisibility = (): TVisiblity => {
```
farzin-deriv, line 9: @michio-rework Just a suggestion, I think something like `useScreenSize` would be a better name considering the return values of the hook.

src/store/index.tsx
```diff
@@ -75,6 +79,7 @@ export const DerivProvider = ({ children }: DerivProviderProps) => {
                 deriv_api,
                 show_non_eu_popup,
                 setShowNonEuPopup,
+                breakpoints,
```
farzin-deriv, line 82: @michio-rework Why would we need to keep this in the context? 🤔

src/components/containers/visibility.tsx
```diff
@@ -78,3 +79,19 @@ export const Mobile = ({
         <MobileLayer breakpoint={breakpoint_size}>{children}</MobileLayer>
     )
 }
+
+type TVisiblityProps = {
+    config: Partial<TVisiblity>
+    children?: ReactNode
+}
+
+export const Visibility = ({ children, config }: TVisiblityProps) => {
```
farzin-deriv, line 88: @michio-rework If I got it correctly this component is going to replace other components in this file such as `Mobile` and `Desktop`, If that’s the case then I would suggest marking those components as deprecated to warn developers when they are using them in new places and inform them about the new component.

src/components/hooks/use-breakpoints.tsx
michio-rework: `large` it is :sunglasses:

src/components/hooks/use-visibility.tsx
michio-rework: cool, thanks :pray:

src/components/hooks/use-visibility.tsx
michio-rework: another thing is, we will create a separate `Context` for the screen sizes. DerivStore is handling too much.

src/components/hooks/use-breakpoints.tsx
michio-rework: oh yeah, we surly add those, just wanted to see the feedback, we're starting as simple as possible, but yeah will them thanks.

src/pages/trade-types/cfds/_start-trading.tsx
michio-rework: update, thanks :pray:

src/store/index.tsx
michio-rework: we're gonna have a separate `Context` for this one, but for this step we're gonna use the `DerivStore` for now. 
the reason behind the using `context` is mentioned in above comments.

src/components/hooks/use-visibility.tsx
michio-rework: yeah you're right, i was struggling with the name, thanks :pray:

src/components/containers/visibility.tsx
michio-rework: yeah we will, it's gonna be another task. 
making these components deprecated will cause confusion right now. 
will do it in the next step ( iteration ).

habib-deriv (approved):

PR #3871, opened 2022-11-22
Mitra/81285/show whatsapp to all countries

mitra-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/components/layout/nav/nav-partner.tsx
```diff
@@ -12,16 +11,13 @@ type NavPartnersProps = {
...
     return (
-        <>
-            <PartnerWrapper>
-                <MainNav />
-                <PartnerNavigationBarWrapper>
-                    <NavPartnerDesktop hide_login_signup={hide_login_signup} />
-                    <NavPartnerMobile hide_login_signup={hide_login_signup} />
-                </PartnerNavigationBarWrapper>
-            </PartnerWrapper>
-            <CFDWarning />
```
mitra-deriv, line 23: Not sure why we have been rendering `CFDWarning` separately again in `NavPartners` and `NavSecurity`, We have just removed them and rendering it only once in the `LayoutOverlay` instead.

src/components/custom/cookie-banner.tsx
```diff
@@ -115,28 +89,44 @@ const LinkText = styled(LocalizedLinkText)`
...
+const CookieBanner = () => {
+    const cookie = useCookieBanner()
+    const [is_visible, setIsVisible] = React.useState(false)
+
+    React.useEffect(() => {
+        if (is_visible !== cookie.should_show) {
+            if (cookie.should_show) {
+                setIsVisible(cookie.should_show)
+            } else {
+                setTimeout(() => setIsVisible(cookie.should_show), 200)
```
mitra-deriv, line 101: In order to show a fade out animation when the dialog is closing we have to duplicated the `should_show` state and add a delay to it, Later we can improve this.

konstantinosG-deriv: So `setIsVisible` here is false in both **if** statements?

yashim-deriv (changes requested):
Can we remove the `is_visible` and `setVisible` and use the `should_show` state directly? You can actually offset this using the `animation_delay` css rule:

https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay#:~:text=The%20animation%2Ddelay%20CSS%20property,and%20partway%20through%20the%20animation.

src/components/custom/cookie-banner.tsx
mitra-deriv: `is_visible` is just a copy of `should_show`, If `should_show` is `true` then `is_visible` is also `true` and if `should_show` is `false` then `is_visible` is also `false`, The only thing is that we are adding a delay while setting `is_visible` to false and wait for the animation to finish.

mitra-deriv:
> Can we remove the `is_visible` and `setVisible` and use the `should_show` state directly? You can actually offset this using the `animation_delay` css rule:
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay#:~:text=The%20animation%2Ddelay%20CSS%20property,and%20partway%20through%20the%20animation.

----

@yashim-deriv Not sure how `animation_delay` can help here 🤔

In the [previous implementation](https://github.com/binary-com/deriv-com/pull/3871/files#diff-27e90ba3c5be58c9805bc09efd53f7e959b2bd1eaee812489e4b073939fc1660L120) we were keeping the element in the DOM and just hide it by setting the opacity to zero, With the [current implementation](https://github.com/binary-com/deriv-com/pull/3871/files#diff-27e90ba3c5be58c9805bc09efd53f7e959b2bd1eaee812489e4b073939fc1660R106) I'm trying to remove the element from the DOM after animation is finished because there is no need to keep it in the DOM anymore.

Maybe I'm missing something, Would you please let me know how can I improve and achieve this by using `animation_delay` 🙏

yashim-deriv (dismissed):

michio-rework (approved):

habib-deriv (approved):

PR #3898, opened 2022-11-29
yashim/fix: allow_tracking event on non eu-uk countries

yashim-deriv:
Changes:
-  Removed double check. `is_tracking_accepted` is not supposed to be checked for non-eu and it is already checked in `allow_tracking`

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


prince-deriv (dismissed):

src/components/layout/layout.tsx
```diff
@@ -187,14 +187,12 @@ const Layout = ({
             if (allow_tracking) {
                 window.onload = () => {
                     window.setTimeout(() => {
```
habib-deriv, line 189: Hi @yashim-deriv  just a question, do we need the setTimeout here, as it already has the dependencies?

yashim-deriv: This is from an older implementation. I think the reasoning is to fire this event after the `page_load` event defined in the `gatsby-browser.js`

michio-rework (approved):

habib-deriv (dismissed):
LGTM

habib-deriv (approved):

PR #3934, opened 2022-12-09
Prince/Update translation workflow security

prince-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/.github/workflows/translation-delete.yml
+++ b/.github/workflows/translation-delete.yml
@@ -12,7 +12,11 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: 'Branch [${{ github.event.workflow_run.head_branch }}]'
-              run: echo 'Deleting Translation Branch ${{ github.event.workflow_run.head_branch }}'
+              run: |
+                BRANCH='${{ github.event.workflow_run.head_branch }}'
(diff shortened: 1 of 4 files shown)


harish-deriv:
@prince-deriv 

Can the below be implemented additionally like in the below reference link snippet instead of referencing the ${{ github.event.workflow_run.head_branch }}' directly in echo.
 It's not a good security practise to inject a variable directly inside the '. Referencing the below way has string escaping built in. This would mean to have something like this

`env:   
      BRANCH_NAME: ${{ github.event.workflow_run.head_branch }}   
run: |
echo "pUshIng strings and translations from $BRANCH_NAME"
`
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#good-practices-for-mitigating-script-injection-attacks


Also having the CLEAN_BRANCH sanitization would be additional security protection after the initial pre push security checks.

PR #3948, opened 2022-12-14
refactor: semantic rules in language dropdown

ashraf-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/elements/typography.ts
+++ b/src/components/elements/typography.ts
@@ -256,31 +256,41 @@ export const TextContainers = styled(({ as = 'h2', children, ...props }: HeaderP
     font-weight: ${(props) => props.weight || 'bold'};
     font-size: ${(props) => {
         if (props.size) return props.size
-        else
-            return (
-                ElementStyleAttributes[props.type] &&
-                ElementStyleAttributes[props.type]['font_size']
-            )
+        else {
+            if (typeof props.type === 'string') {
+                return (
+                    ElementStyleAttributes[props.type] &&
+                    ElementStyleAttributes[props.type]['font_size']
+                )
+            }
+        }
     }};
     line-height: ${(props) => {
-        return (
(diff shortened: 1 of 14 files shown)


src/components/elements/typography.ts
```diff
@@ -92,6 +92,11 @@ export const BaseElement = css<BaseElementProps>`
     text-align: ${(props) => props.align || 'start'};
     padding: ${(props) => props.padding || ''};
     color: ${({ color }) => (color ? `var(--color-${color})` : 'var(--color-black-3)')};
+
+    &[data-color='red'] {
+        color: var(--color-red);
```
ali-hosseini-fs, line 97: why cant we use the color prop that used in line 94?

ashraf-deriv: Because of w3c is complaining about the `color` attribute , we are avoid using that.

ali-hosseini-fs: so in that case whats the point of line 94?
i mean i'm thinking that styled components doesn't add all of the attrs on the element unless we add them manually, is that right?

ashraf-deriv: @ako Yes, That is right. 
In this particular PR we are reducing the number of w3c errors on `Attribute color not allowed on element [p](https://html.spec.whatwg.org/multipage/#the-p-element) at this point.`  
and line 94 still valid but with the cost of w3c errors. if we want to remove the all errors from w3c on this we have to get rid of the `color` attribute from <Header> component and use data- attributes instead.

PR #4029, opened 2023-01-11
NikitK/ Signup not working after opening login page on Safari browser

NikitK-deriv:
Changes:

-    init reopen ws if its closed

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/signup/index.tsx
+++ b/src/pages/signup/index.tsx
@@ -67,7 +67,6 @@ const NewSignup = () => {
     if (WS?.isConnectionClosed()) {
         WS.openHandler()
     }
-    console.log(WS?.isConnectionClosed())
 
     const updateSubmitState = (
         status_arg: React.SetStateAction<string>,
--- a/src/pages/signup/index.tsx
+++ b/src/pages/signup/index.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react'
 import styled from 'styled-components'
+import { useDerivApi } from '../../components/hooks/use-deriv-api'
 import Signup, { Appearances } from 'components/custom/signup'
 import Layout from 'components/layout/layout'
 import { localize, WithIntl } from 'components/localization'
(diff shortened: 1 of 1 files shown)


src/pages/signup/index.tsx
```diff
@@ -61,6 +62,11 @@ const StyledGraph = styled.img`
 const NewSignup = () => {
     const [submit_state, setSubmitState] = useState('')
     const [email, setEmail] = useState('')
+    const { WS } = useDerivApi()
+
+    if (WS?.isConnectionClosed()) {
+        WS.openHandler()
+    }
```
farzin-deriv, line 69: @NikitK-deriv How about we move this to the `useDerivApi` hook itself and whenever the `send` is called we can check the connection status and reconnect, This will also solve the issue for other places if exists 🤔

prince-deriv: @farzin-deriv  that sounds good, however I was thinking if there's a chance we would want to close the websocket for good, in that case it will keep the connection alive. For now I couldn't think of a reason we would want the ws connection close for good.

farzin-deriv: @prince-deriv Yeah that’s a good point, Unfortunately I'm not very familiar with the `deriv-com` codebase to know all the use cases 😅 Maybe someone from the team can comment on this 🙇🏻

NikitK-deriv: @farzin-deriv As far as I know we use in deriv.com ping-pong method to keep connection alive. Already implemented by me and Michio. 

This realization only reach an aim to reopen connection due the specific safari browser behavior. When we use external link and return back to deriv.com. Chrome just download page and everything works, but safari return back to page with already closed websocket. 

Thanks for comment 🙇

farzin-deriv: @NikitK-deriv My concern is that now we are making the Sign Up page responsible for socket connection which is not correct, So let's say we face the issue on some other page as well then we will also need to make those pages responsible for the same thing and have this patch everywhere, That’s why I'm suggesting to move it in the hook which is responsible for socket.

It's just a suggestion tho, We can ignore it 🙇🏻

PR #4060, opened 2023-01-18
NikitK / Task - Refactor check-email page [Translation Key Extraction]

NikitK-deriv:
Changes:

-   TKE check-emails page

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/check-email/_icon-grid.tsx
+++ b/src/pages/check-email/_icon-grid.tsx
@@ -2,13 +2,14 @@ import React from 'react'
 import styled from 'styled-components'
 import device from 'themes/device'
 import { Text } from 'components/elements'
-import { localize } from 'components/localization'
+import { Localize, localize } from 'components/localization'
 // Icons
 import Spam from 'images/svg/check-email/spam.svg'
 import Typo from 'images/svg/check-email/typo.svg'
 import WorkEmail from 'images/svg/check-email/work-email.svg'
 import Firewalls from 'images/svg/check-email/firewalls.svg'
 import { Container } from 'components/containers'
+import { TString } from 'types/generics'
 
 const Grid = styled(Container)`
     display: grid;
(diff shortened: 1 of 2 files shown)


src/pages/check-email/_icon-grid.tsx
```diff
@@ -80,7 +82,7 @@ export const IconGrid = () => (
             <GridCol key={`key-${index}`}>
                 <Img src={item.Icon} alt={item.alt} />
                 <StyledText color="black-3" lh="1.55" mt="0.8rem">
-                    {item.content}
+                    <Localize translate_text={item.content} />
```
prince-deriv, line 85: Why do we still need to localize this? We have implemented the new translation key in the declaration of these items.

niloofar-deriv: @prince-deriv We still need to use this component for extracting our translations.
In the previous version we couldn't use dynamic content inside our `<Localize />` component so we had to pass the localize inside our data_array but right now we can pass dynamic content to our `<Localize />` component by using `_t_` pre/postfix

cc: @mohsen-deriv

niloofar-deriv:
```
export const Localize = ({ translate_text, values, components }: LocalizeProps) => {
    if (typeof translate_text === 'string' && translate_text?.includes('_t_')) {
        // Since I know the indices I wanna remove, substring is used instead of regex
        const actual_key = translate_text.substring(3, translate_text.length - 3)
        return <Trans defaults={actual_key} values={values} components={components} />
    }
    return <Trans defaults={translate_text} values={values} components={components} />
}
```

mohsen-deriv: we have to use the `Localize` component no matter what, it taps into translation context and gets the actual translation for the `key`, the POC only handles better key extraction and better usage of `Localize` component or `localize` function.

niloofar-deriv (approved):

prince-deriv (approved):

mohammad-hashemi-deriv (approved):

PR #4096, opened 2023-01-25
Niloofar | Refactoring translations (localize / Localize) | FE Core Sprints [09 Jan - 29 Jan]

niloofar-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [x] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/academy/blog/index.tsx
```diff
@@ -51,6 +52,12 @@ type ArticlesPageProps = {
 
 export type ArticleDataType = AllArticlesQuery['directus']['blog']
 
+const title_text: TString = '_t_Articles, trading guide and resources | Deriv_t_'
+const description_text: TString =
+    "_t_If you are looking for trading guide or tutorials, visit Deriv's trading academy and learn how to trade online._t_"
+const blog_text: TString = '_t_Blog_t_'
+const least_articles_text: TString = '_t_The latest articles and resources_t_'
```
farzin-deriv, line 59: @niloofar-deriv Just for my understanding, What’s the purpose of doing this? What’s the problem we are trying to solve? 🤔

niloofar-deriv: Let's huddle about it

farzin-deriv: As discussed in the call, We will modify the `extract-translations` script to accept both legacy and the new way of passing text to localize components so we don’t have to extract the texts into const variables and we can also update the type of localize component prop to complain about `_t_foo_t_` pattern so after that devs only need to fix TS errors and don’t need to know how the script works.

@niloofar-deriv @mohsen-deriv

mohsen-deriv (dismissed):

PR #4244, opened 2023-02-24
Mitra/Pr 88609 remove create demo account when client come from deriv go

mitra-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/pages/markets/components/sections/_hero_derived_fx.tsx
```diff
@@ -108,7 +110,9 @@ export const DerivedFXHero = ({ title, description }: MarketProps) => {
                 <MarketSubHeader color="white" weight="normal" align="center">
                     <Localize translate_text={description} />
                 </MarketSubHeader>
-                {is_logged_in ? (
+                {is_deriv_go ? (
+                    <></>
+                ) : is_logged_in ? (
```
farzin-deriv, line 115: @mitra-deriv Avoid using nested ternary operator.

```tsx
{is_logged_in && !is_deriv_go && <GetTrading />}
{!is_logged_in && !is_deriv_go && <CreateFreeDemoAccount />}
```

src/common/utility.ts
```diff
@@ -424,3 +424,8 @@ export const handleRowRedirect = (residence: string, current_client_country: str
 export const isLocalhost = () => !!(isBrowser() && process.env.NODE_ENV === 'development')
 
 export const isTestlink = () => !!(isBrowser() && window.location.hostname.includes('binary.sx'))
+
+export const getDateFromToday = (num_of_days: number) => {
```
farzin-deriv, line 428: @mitra-deriv Try to write unit test for the newly added utility function.

src/pages/markets/components/sections/_hero_derived_fx.tsx
```diff
@@ -108,7 +110,9 @@ export const DerivedFXHero = ({ title, description }: MarketProps) => {
                 <MarketSubHeader color="white" weight="normal" align="center">
                     <Localize translate_text={description} />
                 </MarketSubHeader>
-                {is_logged_in ? (
+                {is_deriv_go ? (
```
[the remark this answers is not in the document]
mitra-deriv: @fasihali-deriv Thank you so much for the comment, As we discussed in person we have another condition so this approach won’t work :( 🙏  Thanks

src/common/utility.ts
mitra-deriv: This change was removed from master in resolving conflicts I kept it by mistake, thanks for pointing it out, Removed it as it's not part of my changes.

habib-deriv (approved):
LGTM

src/common/utility.ts
```diff
@@ -424,3 +424,15 @@ export const handleRowRedirect = (residence: string, current_client_country: str
 export const isLocalhost = () => !!(isBrowser() && process.env.NODE_ENV === 'development')
 
 export const isTestlink = () => !!(isBrowser() && window.location.hostname.includes('binary.sx'))
+
+export const PlatformQueryParam = () => {
```
farzin-deriv, line 428: @mitra-deriv Is there a reason for not using the `usePlatformQueryParam` hook and creating this function instead? Is it related to server-side rendering? 🤔

mitra-deriv: Not sure whats the issue with `gatsby query param` but when refreshing the page we have a flickering (some delay) before the sections gets hidden, so I used this function and set it true so we don’t have that flickering and it worked, if there is any better way to handle the flickering I will be happy to implement it 🙏

farzin-deriv: @mitra-deriv I wouldn't suggest duplicating the code as we will need to maintain both places and over time each function will go in a different direction.

The flickering is probably happening because `useQueryParam` at first returns `undefined` and then gets updated with the value and causes rerendering, Maybe you can avoid this by adding an extra `state` and a `useEffect` in the `usePlatformQueryParam` hook and using a default value. (Let me know if you need help on this)

Although we already have this behavior in production, For example, if you go to https://deriv.com/?platform=p2p you can see that the header disappears after being rendered first, So maybe you can ignore it for now.

mitra-deriv: Thanks for the suggestion I will try to fix this in a new PR and resolve this conversation once done 🙏 

cc: @habib-deriv

mitra-deriv: Hi @farhan-nurzi-deriv , Once you applied the fix in your PR help to mention the PR number here so we can resolve this conversation.

Thank you so much for the help  🙏

farhan-nurzi-deriv: This issue is resolved here https://github.com/binary-com/deriv-com/pull/4310 🙏

PR #4346, opened 2023-03-16
Fasih/: fixed user prefered language issue

fasihali-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


gatsby-browser.js
```diff
@@ -20,6 +22,26 @@ const checkDomain = () => {
     )
 }
 
+const updateURLAsPerUserLanguage = () => {
```
farzin-deriv, line 25: @fasihali-deriv Would you please extract this function into a utility file and write unit tests for it 🙏🏻

farzin-deriv (changes requested):

gatsby-browser.js
```diff
@@ -20,6 +22,26 @@ const checkDomain = () => {
...
+    const current_path = window.location.pathname
+    const current_hash = window.location.hash
+    const paths = current_path.split('/')
+    const first_path = paths[1]
+    const has_language_in_url = first_path in language_config
```
yashim-deriv, line 30: If u use this regex `\/(?=(en|es|fr|id|it|pl|pt|ru|th|vi|tr|zh-cn|zh-tw|bn|ar|ko|ach|de)\/)\1` u can match all supported languages in language config while supporting any type of URLs

yashim-deriv: If you need it to be dynamic u can load the keys only from language_config and serialize it into a string. See screenshot below, it should support all language code
Screenshot 2023-03-16 at 12 11 11 PM

yashim-deriv (commented):
You should split the function into three:
- One to check if the language code is valid
- One to store language into the cookies
- One to update the URL

Then you should write a test for it based on what farzin recommended

gatsby-browser.js
fasihali-deriv: Hi @yashim-deriv Thanks for mentioning this regex concept over here. I can update my code according to your suggestion but I have a question 👀  . 

What is benefit behind that?
To achieve this thing I need to Serialize that Object then using that I will have dynamic regex and then I will compare it. 
Why not just `in` ? 👀

gatsby-browser.js
habib-deriv: Hi @fasihali-deriv @farzin-deriv  let's move the function to utility. As tests are not yet setup, we are good to go for QA after this.

habib-deriv (approved):
LGTM

habib-deriv (approved):

PR #4369, opened 2023-03-21
The font size of caption in Deriv IB Programme should be bigger

farhan-nurzi-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/pages/partners/affiliate-ib/_minimum-volume.tsx
+++ b/src/pages/partners/affiliate-ib/_minimum-volume.tsx
@@ -39,7 +39,7 @@ const StyledSection = styled(SectionContainer)`
         padding: 40px 0;
     }
 `
-const TitleWrapper = styled(Flex)`
+const TitleWrapper = styled.div`
     max-width: 120rem;
     margin: 0 auto 2.4rem;
 
@@ -50,13 +50,9 @@ const TitleWrapper = styled(Flex)`
 
     @media ${device.tabletL} {
         margin: 0 auto 16px;
-
-        ${Header} {
-            font-size: 20px;
-            line-height: 25px;
-        }
     }
 `
+
 const StyledHeader = styled(Header)`
     max-width: 1200px;
     margin: 0 auto 24px;
@@ -67,6 +63,19 @@ const StyledHeader = styled(Header)`
         margin: 0 auto 16px;
     }
 `
+
+const SubtitleHeader = styled(Header)`
+    line-height: 36px;
+
+    @media ${device.laptopL} {
+        width: 100%;
+    }
+    @media ${device.tabletL} {
+        line-height: 24px;
+        font-size: 16px !important;
+    }
+`
+
 const ExampleWrapper = styled(Flex)`
     max-width: 120rem;
     margin: 0 auto;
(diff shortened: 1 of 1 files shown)


src/pages/partners/affiliate-ib/_minimum-volume.tsx
```diff
@@ -67,6 +63,19 @@ const StyledHeader = styled(Header)`
...
+
+const SubtitleHeader = styled(Header)`
+    line-height: 36px;
+
+    @media ${device.laptopL} {
+        width: 100%;
+    }
+    @media ${device.tabletL} {
+        line-height: 24px;
+        font-size: 16px !important;
```
mohammad-hashemi-deriv, line 75: Can we avoid using `!important` here?

src/pages/partners/affiliate-ib/_minimum-volume.tsx
```diff
@@ -67,6 +63,19 @@ const StyledHeader = styled(Header)`
         margin: 0 auto 16px;
     }
 `
+
+const SubtitleHeader = styled(Header)`
+    line-height: 36px;
```
NikitK-deriv, line 68: line-height exist when we passing `type="sub-section-title"`
Should we have different line height here?

src/pages/partners/affiliate-ib/_minimum-volume.tsx
```diff
@@ -67,6 +63,19 @@ const StyledHeader = styled(Header)`
...
+
+    @media ${device.laptopL} {
+        width: 100%;
+    }
+    @media ${device.tabletL} {
+        line-height: 24px;
+        font-size: 16px;
```
NikitK-deriv, line 75: This property exist too
Better approach it find the figma's design, check px of title and pass type equal for desktop 
Screenshot 2023-03-22 at 09 18 50

src/pages/partners/affiliate-ib/_minimum-volume.tsx
```diff
@@ -39,7 +39,7 @@ const StyledSection = styled(SectionContainer)`
         padding: 40px 0;
     }
 `
-const TitleWrapper = styled(Flex)`
+const TitleWrapper = styled.div`
```
aswathy-deriv, line 42: Any reason why flex was removed to div

farhan-nurzi-deriv: Because the texts alignment wasn't centered with flex, and i use div to follow the style used in the section above it which also use div. Also, without flex I don't need to pass extra props in titlewrapper, but still get the expected behaviour.

prince-deriv: why not keep the flex and add the alignment properties instead? Later this might change again and will require alignment control.

NikitK-deriv (approved):

mohammad-hashemi-deriv (approved):

PR #4408, opened 2023-03-24
fix: store selected language in cookies

Nuzhy-Deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/common/utility.ts
```diff
@@ -403,7 +403,8 @@ export const updateURLAsPerUserLanguage = () => {
     const paths = current_path.split('/')
     const first_path = paths[1]
     const has_language_in_url = first_path in language_config
-    has_language_in_url && Cookies.set('user_language', first_path)
+    document.cookie =
+        has_language_in_url && 'user_language=' + first_path + ';domain=.' + getDomain() + ';path=/'
```
prince-deriv, line 407: Is there a library that can handle this or can we find any existing helpers that can do thesame thing? Otherwise let's create our own in the `helpers` directory to keep this clean.

src/common/utility.ts
```diff
@@ -403,7 +403,8 @@ export const updateURLAsPerUserLanguage = () => {
...
     const user_language = Cookies.get('user_language') || 'en'
```
NikitK-deriv, line 408: In case when we are not set Cookie, we have to rewrite it on:
```suggestion
    const user_language = document.cookie || 'en'
```

NikitK-deriv (commented):
By the way, any idea what is it?

Screenshot 2023-03-24 at 13 35 33

src/common/utility.ts
```diff
@@ -403,7 +403,8 @@ export const updateURLAsPerUserLanguage = () => {
     const paths = current_path.split('/')
     const first_path = paths[1]
     const has_language_in_url = first_path in language_config
-    has_language_in_url && Cookies.set('user_language', first_path)
+    document.cookie =
```
mohammad-hashemi-deriv, line 406: By this approach I think we are creating a `cookie` without any name right?
If `has_language_in_url` would be `false` we will store a cookie with `false` value without any name.
Let's avoid this.

Nuzhy-Deriv:
Closing this PR to avoid conflicts

PR #4956, opened 2023-06-07
Update live pricing table using firebase sdk

habib-deriv:
## Overview

This pull request introduces Firebase into the application for handling data fetching from the 'deriv-static-pricingfeed' database. This change optimizes our current data fetching method and leverages Firebase's real-time database features for fetching live pricing data, making adjustments to the `LiveMarketTable` component. 

### Data Fetching Hook

- Introduced a new custom hook named `usePricingData`. This hook is responsible for initialising the Firebase app, getting the database, and setting up a subscription to the specific market data based on the user's region. It then provides the real-time data or error through its return object.

### Live Market Table Component

- Adjusted `LiveMarketTable` component to use `usePricingData` hook for fetching market data.
- Removed the old method of fetching data every 10 seconds with `setInterval`. Instead, we now have real-time data fetching with Firebase.
- Simplified code by removing redundant state variables and effects.

### Style Changes

- Removed `.visible-larger-than-phone` class from `visibility.scss` as it was not being used in the project.

## Impact

This change enhances the efficiency of our live pricing feed by leveraging Firebase's real-time capabilities. It also reduces the complexity in our `LiveMarketTable` component.

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/hooks/use-firebase-fetch-data.tsx
+++ b/src/components/hooks/use-firebase-fetch-data.tsx
@@ -1,0 +1,33 @@
+import { useState, useEffect } from 'react'
+import { initializeApp } from 'firebase/app'
+import { getDatabase, ref, onValue } from 'firebase/database'
+import { firebaseConfig } from 'common/constants'
+import useRegion from 'components/hooks/use-region'
+
+const useFetchFirebaseData = () => {
+    const [data, setData] = useState(null)
+    const [error, setError] = useState(null)
+    const { is_eu } = useRegion()
+
+    useEffect(() => {
+        const app = initializeApp(firebaseConfig)
+        const db = getDatabase(app)
+
+        const commoditiesRef = is_eu ? ref(db, 'eu/market') : ref(db, 'row/market')
+        const unsubscribe = onValue(
+            commoditiesRef,
+            (snapshot) => {
+                setData(snapshot.val())
+            },
+            (err) => {
+                setError(err)
+            },
+        )
+
+        return unsubscribe
+    }, [is_eu])
+
+    return { data, error }
+}
+
+export default useFetchFirebaseData
--- a/src/features/pages/home/live-pricing/table-component/live-market-table.tsx
+++ b/src/features/pages/home/live-pricing/table-component/live-market-table.tsx
@@ -14,11 +14,36 @@ import Link from 'features/components/atoms/link'
 import { Localize } from 'components/localization'
 import usePricingFeed from 'components/hooks/use-pricing-feed'
 import Typography from 'features/components/atoms/typography'
+import InitialLoader from 'components/elements/dot-loader'
 
 export type TLiveMarketTableProps = {
     selected_market: TAvailableLiveMarkets
     link_to: string
 }
+
+const Loader = (
+    <Flex.Box justify="center">
+        <InitialLoader
+            style={{
+                padding: '40px 0',
+                marginInlineStart: '45%',
+            }}
+        />
+    </Flex.Box>
+)
+const ErrorView = (
+    <Flex.Box justify="center" mt="5x">
+        <Typography.Paragraph
+            align="left"
+            padding_block="3x"
+            padding_inline="6x"
+            font_family="UBUNTU"
+            textcolor="brand"
+        >
+            <Localize translate_text="_t_There was an error fetching the live pricing data_t_"></Localize>
+        </Typography.Paragraph>
+    </Flex.Box>
+)
 
 const LiveMarketTable = ({ selected_market, link_to }: TLiveMarketTableProps) => {
     const [error, rawMarketsData] = usePricingFeed()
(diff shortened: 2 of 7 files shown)


src/components/hooks/use-firebase-fetch-data.tsx
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from 'react'
+import { initializeApp } from 'firebase/app'
+import { getDatabase, ref, onValue } from 'firebase/database'
+import { firebaseConfig } from 'common/constants'
+import useRegion from 'components/hooks/use-region'
+
+const useFetchFirebaseData = () => {
```
markwylde-deriv, line 7: Can we rename this to?

```suggestion
const usePricingFeed = () => {
```

src/components/hooks/use-firebase-fetch-data.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const [data, setData] = useState(null)
+    const [error, setError] = useState(null)
+    const { is_eu } = useRegion()
+
+    useEffect(() => {
+        const app = initializeApp(firebaseConfig)
+        const db = getDatabase(app)
+
+        const commoditiesRef = is_eu ? ref(db, 'eu/market') : ref(db, 'row/market')
```
markwylde-deriv, line 16: Do you think we should switch the turnery to inside the ref?

```suggestion
        const commoditiesRef = ref(db, is_eu ? 'eu/market' : 'row/market')
```

src/components/hooks/use-firebase-fetch-data.tsx
```diff
@@ -0,0 +1,33 @@
...
+            },
+            (err) => {
+                setError(err)
+            },
+        )
+
+        return unsubscribe
+    }, [is_eu])
+
+    return { data, error }
```
markwylde-deriv, line 30: Can we return this as an array, with the `error` first. This way, we are forced/reminded to handle the error.

```suggestion
    return [error, data]
```

src/features/pages/home/live-pricing/table-component/live-market-table.tsx
```diff
@@ -10,58 +10,33 @@ import { TAvailableLiveMarkets, TMarketData } from '../types'
...
 
 export type TLiveMarketTableProps = {
     selected_market: TAvailableLiveMarkets
     link_to: string
 }
 
 const LiveMarketTable = ({ selected_market, link_to }: TLiveMarketTableProps) => {
-    const { is_eu } = useRegion()
-    const [initial_loaded, setInitialLoaded] = useState(false)
+    const firebaseData = useFetchFirebaseData()
```
markwylde-deriv, line 23: If we make the change I suggested above, we can now return the 

```suggestion
    const [error, pricingData] = useFetchFirebaseData()
```

Then we can handle the `error` by maybe displaying an error, unless there is already a design for this.

src/features/pages/home/live-pricing/table-component/live-market-table.tsx
```diff
@@ -10,58 +10,33 @@ import { TAvailableLiveMarkets, TMarketData } from '../types'
...
-            getData()
-            setInitialLoaded(true)
-        }
-        intervalRef.current = setInterval(getData, 10000)
-        return () => {
-            clearInterval(intervalRef.current)
-        }
-    }, [getData, initial_loaded])
-
     const updateData = useCallback(() => {
```
markwylde-deriv, line 34: I don't think we need to do this as a callback any more.

Maybe instead the mutated data can be a `useMemo`?

Then we can remove these lines from below:

```javascript
    useEffect(() => {
        updateData()
    }, [updateData])
```

markwylde-deriv (changes requested):

src/features/pages/home/live-pricing/table-component/live-market-table.tsx
habib-deriv: Yeah that's right useMemo works for us, but to load the function and not have multiple render kept the useEffect hook

markwylde-deriv: Hmm, do you think it needs to be a function though. It seems like the only output of that `updateData` function is to run `setMarketsData`, therefore setting `marketsData` to the new value.

I was wondering, can we not just do `const marketsData = useMemo(() => {...`

markwylde-deriv (approved):
Looks really good. Nice one Habib 👍

markwylde-deriv (approved):
Looks really good. Nice one Habib 👍

src/features/pages/home/live-pricing/table-component/live-market-table.tsx
```diff
@@ -92,6 +56,21 @@ const LiveMarketTable = ({ selected_market, link_to }: TLiveMarketTableProps) =>
...
+        return (
+            <Flex.Box justify="center" mt="5x">
+                <Typography.Paragraph
+                    align="left"
+                    padding_block="3x"
+                    padding_inline="6x"
+                    font_family="UBUNTU"
+                    textcolor="brand"
+                >
+                    There was an error fetching the live pricing data
```
NikitK-deriv, line 69: Should we translate error message? And probably make it red

NikitK-deriv (approved):
Good idea to use sdk. I thought we didn't use it by some reasons

src/features/pages/home/live-pricing/table-component/live-market-table.tsx
habib-deriv: Thanks @NikitK-deriv , it's better to have this translated. textcolor="brand" is making it red.

PR #4957, opened 2023-06-08
fix: landing page issues

Nuzhy-Deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/common/utility.ts
```diff
@@ -415,15 +415,18 @@ export const updateURLAsPerUserLanguage = () => {
...
 
-    const language = has_language_in_url ? first_path : user_language
-
     if (!has_language_in_url && user_language === 'en') return
     if (first_path === user_language) return
+    if (first_path !== user_language) {
+        Cookies.remove('user_language')
+        setCookiesWithDomain('user_language', first_path)
+        return
+    }
```
meenakshi-deriv, line 424: why are we using two if statements? can we do it in if else

Nuzhy-Deriv: we can't use `else return` in this case. if we use, the codes below this condition will be unreachable

src/common/utility.ts
```diff
@@ -415,15 +415,18 @@ export const updateURLAsPerUserLanguage = () => {
     }
     const user_language = Cookies.get('user_language') || 'en'
 
-    const language = has_language_in_url ? first_path : user_language
-
     if (!has_language_in_url && user_language === 'en') return
     if (first_path === user_language) return
+    if (first_path !== user_language) {
```
meenakshi-deriv, line 420: @Nuzhy-Deriv why are we using two if statements? Can we do it in `if else`?
 ```
if (first_path !== user_language) {
        Cookies.remove('user_language')
        setCookiesWithDomain('user_language', first_path)
        return
    }
    else return
```

File: src/pages/landing/ebooks/components/_get-ebook.tsx [hunk 3]
```diff
@@ -343,7 +343,7 @@ const GetEbook = ({ color = 'var(--color-white)', ebook_utm_code, onSubmit }: Ge
                             components={[
                                 <StyledLocalizedLink
                                     key={0}
-                                    to="/tnc/security-and-privacy.pdf"
+                                    href="/tnc/security-and-privacy.pdf"
```
NikitK-deriv, line 346: Probably we have to make `is_eu` condition because we have different rules for them. Left comment in card, let's wait QA decision

NikitK-deriv (approved):

mitra-deriv (approved):

PR #4969, opened 2023-06-12
Shayan/ FEQ-286/ fixed jest errors and refactored updated sample test 

shayan-deriv:
Changes:
Since the codebase has been updated the sample test file failed and to fix the issue I've updated jest version and also fixed the issue on the existing sample file but since we are going to migrate all components from styled component to sass, I've created a new test file for checkbox component to make sure there is no issue with the jest setup.

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/features/components/atoms/input/checkbox/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+describe('Test name', () => {
+    test('should be true', () => {
+        render(
+            <CheckboxInput>
+                {' '}
+                <div>shayan</div>
+            </CheckboxInput>,
+        )
+        expect(screen.getByText('shayan')).toBeInTheDocument()
```
mohsen-deriv, line 13: please add assertions for `checked` and `unchecked` status of the component, thanks :pray :fire:

mohsen-deriv (changes requested):
can we update the test folders to only `features`? 
and exclude `features/pages` as well?

src/features/components/atoms/input/checkbox/__tests__/index.spec.tsx
shayan-deriv: updated the test. would you please check it again?

svgTransform.js
```diff
@@ -2,7 +2,9 @@
 
 module.exports = {
     process() {
-        return 'module.exports = {};'
+        return {
+            code: `module.exports = {};`,
+        }
```
amir-deriv, line 7: have you tested this with svgs ? in my case it was not working

@shayan-deriv can you try rendering the page component in a test file just to make sure that svg's are getting trasnformed correctly ?

amir-deriv: https://github.com/binary-com/deriv-com/blob/adbd2ae18de53b3b411ccf93bb3ba687c10932b7/__mocks__/svg-mock.js for reference

shayan-deriv: @amir-deriv yes it is working with the new components which are not using styled-component. 

image



but for components which are still using style-component we have a workaround to fix the error.check this file change: 
https://github.com/binary-com/deriv-com/pull/4969/files#diff-fe8049671ea2bf6f8d3af9176022ed50403e26801eb6103eaed1d136a298edc3

but since we are going to remove style-component in the next few months and re-write them using sass , there shouldn't be any worries about this.

ali-hosseini-deriv (approved):

habib-deriv (approved):
LGTM

shayan-deriv:
@ali-hosseini-deriv we can ignore the gitgaurdian errors :pray:

mohsen-deriv (approved):

PR #5186, opened 2023-07-12
Ako/ A/A Test

ali-hosseini-deriv:
Changes:
added A/A test cookies on the project

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


functions/_middleware.js
```diff
@@ -0,0 +1,42 @@
...
+    const cookie = request.headers.get("cookie")
+
+    // client is in the `new` test
+    if (cookie && cookie.includes(`${cookieName}=new`)) {
+        url.search = newHomepagePathName
+        return env.ASSETS.fetch(url)
+    }
+
+    // client is in the `old` test
+    if (cookie && cookie.includes(`${cookieName}=current`)) {
```
habib-deriv, line 22: Hi @ali-hosseini-deriv 
Instead of 'new' and 'current' as cookie name, can we have 'control' as current and 'variant_1' for new. 

Also can we add support for multiple variants?

ali-hosseini-deriv: @habib-deriv 
* For the naming i will do it right now.

* For the multiple variants i think we better focus on two variants at the same time.
the reason is that even now we cannot predict what percentage redirects to variant 1 or 2.
because we are redirecting randomly.
so it can be 40 to 60 percent or even 30 to 70 and so on!
so having more than 2 can cause less useful data.

habib-deriv: For 3 variants the traffic needs to be splitted 33% percent each for variant_a abd variant_b. And control gets 34%. Sure, we can add this as improvement if needed.

chrisgodfrey-deriv:
Looks good, only concern is having slashes in the cookie name: const cookieName = "a/a-test-cookie"

const cookieName = "aa-test-cookie" might be better just to prevent oddness with special characters (possibly) going on.

ali-hosseini-deriv:
> Looks good, only concern is having slashes in the cookie name: const cookieName = "a/a-test-cookie"
> 
> const cookieName = "aa-test-cookie" might be better just to prevent oddness with special characters (possibly) going on.

this is done

habib-deriv (approved):

PR #5310, opened 2023-07-31
feat: use new analytics library

markwylde-deriv:
Use the new [analytics](https://www.npmjs.com/package/@deriv/analytics) library to implement an AB test on `/landing/ebooks/stocks/`.

While nothing can be seen, the tests are on the homepage and the landing page `/landing/ebooks/stocks/`.


src/store/global-provider.tsx
```diff
@@ -1,22 +1,55 @@
...
 import { WebsiteStatusProvider } from './website-status-context'
 import { MediaContextProvider } from 'themes/media'
+import { getClientInformation, getDomain, getLanguage } from 'common/utility'
 
 type GlobalProviderProps = { children: React.ReactNode }
 
+let growthbook
+if (typeof window !== 'undefined') {
+    import('@deriv/analytics').then(({ RudderStack }) => {
+        RudderStack.init()
```
adrienne-deriv, line 15: We don't need to call `RudderStack.init()` anymore, because its automatically initialized once you import it

adrienne-deriv: Screenshot 2023-08-01 at 4 47 49 PM

package.json
```diff
@@ -38,6 +40,7 @@
         "gatsby-plugin-react-helmet": "^5.24.0",
         "gatsby-plugin-react-helmet-canonical-urls": "^1.4.0",
         "gatsby-plugin-robots-txt": "^1.7.1",
+        "gatsby-plugin-rudderstack": "^2.0.0",
```
adrienne-deriv, line 43: I'm not sure if we need this, because the SDK in the `@deriv/analytics` package would handle all of the plugin's described APIs

markwylde-deriv: Nice one @adrienne-deriv I've removed it now. Let's see if it builds.

src/store/global-provider.tsx
markwylde-deriv: Great! Removed it now :)

package.json
```diff
@@ -5,7 +5,9 @@
     "author": "Deriv",
     "dependencies": {
         "@artsy/fresnel": "^3.2.1",
+        "@deriv/analytics": "^1.0.2",
```
farzin-deriv, line 8: @markwylde-deriv Did we publish `@deriv/analytics` on `npm`? 🤔

src/common/constants.ts
```diff
@@ -1,4 +1,5 @@
 const MAP_API_KEY = process.env.GATSBY_MAP_API_KEY
+const GROWTHBOOK_CLIENT_KEY = process.env.GROWTHBOOK_CLIENT_KEY || 'sdk-0aHHaxKhwn0bP1HE'
```
farzin-deriv, line 2: That's a fake fallback API Key, Right? 🤔

package.json
adrienne-deriv, line 8: This needs to be on version `^1.0.3` since the new methods was recently published from `1.0.3` 👀

src/store/global-provider.tsx
```diff
@@ -1,22 +1,61 @@
...
+        const domain = getDomain()
+        const client_information = getClientInformation(domain)
+
+        if (client_information) {
+            RudderStack.identifyEvent(client_information.loginid, {
+                language,
+            })
+        }
+    })
+}
```
farzin-deriv, line 45: @markwylde-deriv Would you please move these into a separate provider file 🙏🏻 I believe `GlobalProvider` is just supposed to wrap the whole app with all the global providers and not contain anything itself 🤔

src/common/constants.ts
habib-deriv, line 2: For gatsby, we'll need to rename the .env variable
```suggestion
const GROWTHBOOK_CLIENT_KEY = process.env.GATSBY_GROWTHBOOK_CLIENT_KEY
```

package.json
adrienne-deriv: https://www.npmjs.com/package/@deriv/analytics

farzin-deriv: So can we remove the local package in deriv-app repository and use the npm package? If we keep it then in less than a month these two packages will be inconsistent and go different path 😅

adrienne-deriv: Yeaah I was planning to do that as well, I've asked ako to create a card for that

markwylde-deriv: Thanks @adrienne-deriv The current published version seems to be 1.0.2:
https://www.npmjs.com/package/@deriv/analytics

Screenshot 2023-08-04 at 08 39 46

src/common/constants.ts
markwylde-deriv: It's our current demo growthbook account. Do we have a way of setting default environment variables for when a new developer checks out the repo?

markwylde-deriv: Thanks @habib-deriv Fixed in 0408a3656ef082debc0dcc1c43072dd105c9d0b1

farzin-deriv: I don't know if we have development env file in deriv-com 🤔 Maybe @habib-deriv can answer that 🙇🏻

adrienne-deriv (approved):
LGTM

src/store/global-provider.tsx
markwylde-deriv: Thanks @farzin-deriv. Absolutely. Fixed in e40166cecbe57dab54be2164ea9b2dfc18d319e0.

src/store/growthbook-context.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { GrowthBook, GrowthBookProvider } from '@growthbook/growthbook-react'
+import { getClientInformation, getDomain, getLanguage } from 'common/utility'
+import { growthbook_client_key } from 'common/constants'
```
farzin-deriv, line 4: Just a suggestion, We can move the key constant to this file and scope it here to have less dependency on other files.

src/store/growthbook-context.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+let growthbook
+if (typeof window !== 'undefined') {
```
farzin-deriv, line 7: Would it be possible to do this in a `useEffect` inside the `DerivGrowthBookProvider` instead of declaring a global variable? 🤔

markwylde-deriv: Yeah nice one. Fixed in 9adf34a6426a604876daf93ef263acb216c85c83.

yashim-deriv (approved):

PR #5347, opened 2023-08-03
Fasih/: :bug  added option to show footer

fasihali-deriv:
Changes:

-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/components/layout/layout.tsx
+++ b/src/components/layout/layout.tsx
@@ -27,6 +27,7 @@ type LayoutProps = {
     is_ppc_redirect?: boolean
     margin_top?: number | string
     type?: string
+    show_footer?: boolean
 }
 
 type MainType = {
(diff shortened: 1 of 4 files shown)


src/components/layout/layout.tsx
```diff
@@ -85,6 +86,7 @@ const Layout = ({
     is_ppc_redirect = false,
     margin_top = '',
     type = 'default',
+    show_footer = true,
```
konstantinosG-deriv, line 89: I think it's true by default so might not need **=true**. Could you check?

habib-deriv (approved):

src/components/layout/layout.tsx
fasihali-deriv: By default it will be `undefined` 
So system will recognise this as false and it will get hide from all the pages where we are not passing any prop.

konstantinosG-deriv (approved):

aswathy-deriv (approved):

PR #5416, opened 2023-08-10
Adrienne / Build to Cloudflare Pages workflow and use caching for dependencies and build output

adrienne-deriv:
Changes:
- Added workflow to delegate Gatsby build to use Github Actions within Github's hosted runners
- Added caching of build output from `npm run build` to leverage on Gatsby's incremental build and speed up build step
- Added caching of dependencies to speed up build step

The `build-to-cloudflare-pages.yml` workflow purpose is to build the current pull request's branch and deploy it to Cloudflare Pages if successful. Caching is introduced to significantly speed up this workflow by leveraging on Gatsby's incremental build, which states that "To be able to use it, you will need to keep .cache and public directories from previous build" https://www.gatsbyjs.com/docs/debugging-incremental-builds/. Once this is introduced, new pull requests and subsequent updates from it will have their builds be completed in approximately 6-8 minutes when the cache is present.
A replica of this cache is present to address the following issue:
- When invalidating the build cache when the `master` branch is updated, the build cache will not be available and will be cleared. Workflows running during this period and fetching from the cleared cache will instead use the replica cache (see diagram below)
- The reason for this is that when a workflow is saving a cache, if at any point that workflow is cancelled or fails, the cache key will be reserved and thus when re-running the invalidation workflow, that cache key is considered 'reserved' and thus is unable to save the new master build, thus one of the solution is to clear the cache and re-set it.
Screenshot 2023-08-10 at 4 27 42 PM

- The cache will be invalidated on a CRON job which runs at `22:00` UTC time, or can be run manually
- diagram 2

The diagram below shows another scenario where, if the `master` branch creates a failing build, that build will NOT be saved in the cache. Other workflows can still enjoy the benefit of incremental build by using the replica cache as a backup while the master cache is cleared. Once a fix is merged into `master` and `sync-build-cache.yml` workflow is re-run successfully, the build cache will be restored.
The invariant is that the build and replica cache will only contain successful builds.
diagram

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


.github/workflows/sync-build-cache.yml
```diff
@@ -0,0 +1,102 @@
...
+                    https://api.github.com/repos/$owner/$repo/actions/caches?key=$cache_key
+
+            - name: Sync backup cache with master cache
+              if: ${{ steps.build-cache.outcome == 'success' }}
+              uses: actions/cache/save@v3
+              with:
+                  key: master-cache-public-replica
+                  path: |
+                      .cache
+                      public
```
balakrishna-deriv, line 102: what is the purpose of `master-cache-public-replica`?

.github/workflows/build-to-cloudflare-pages.yml
```diff
@@ -0,0 +1,123 @@
...
+
+            - name: Build project
+              id: build-project
+              run: npm run build
+
+            - name: Publish to Cloudflare Pages
+              id: publish-to-pages
+              run: |
+                  echo "Installing Wrangler CLI"
+                  npm i -g wrangler
```
balakrishna-deriv, line 86: Can we specify the major version of wrangler so that we don't break the deploy command when there is a new version of wrangler?
```suggestion
                  npm i -g wrangler@3
```

.github/workflows/build-to-cloudflare-pages.yml
```diff
@@ -0,0 +1,124 @@
...
+                  echo "Deploying build to Cloudflare Pages"
+                  directory='public'
+                  projectName='deriv-com'
+                  branch=$GITHUB_HEAD_REF
+
+                  preview_url=$(wrangler pages deploy $directory --project-name=$projectName --branch=$branch > log.txt 2>&1; echo $?)
+                  echo "------"
+                  cat log.txt
+
+                  branchName=$(echo $branch | sed 's/[\/_]/-/g')
```
adrienne-deriv, line 97: Cloudflare Pages formats the default preview link to use `-`, so if a branch has a name like `feq-123/bla-bla`, the preview link will show it as `https://feq-123-bla-bla.deriv-com-playground.pages.dev`

adrienne-deriv: Screenshot 2023-08-10 at 5 54 08 PM

balakrishna-deriv (approved):

.github/workflows/sync-build-cache.yml
```diff
@@ -0,0 +1,102 @@
+name: Sync build cache
+
+permissions:
+    actions: write
+    contents: read
+
+on:
+    schedule:
+        - cron: '0 22 * * *'
```
adrienne-deriv, line 9: This runs everyday on `master` branch around the following time:
- 5:00 AM Malaysia time
- 1:00 AM Dubai time
- 12:00 AM Belarus time

.github/workflows/sync-build-cache.yml
```diff
@@ -0,0 +1,102 @@
...
+    workflow_dispatch:
+
+env:
+    NODE_OPTIONS: '--max-old-space-size=8192'
```
adrienne-deriv, line 13: Same as here, is this space sufficient?

cc: @habib-deriv @balakrishna-deriv

ameerul-deriv (approved):

yashim-deriv (approved):
LGTM

.github/workflows/build-to-cloudflare-pages.yml
```diff
@@ -0,0 +1,118 @@
...
+permissions:
+    actions: read
+    contents: read
+
+on:
+    pull_request:
+        types: [opened, synchronize]
+
+env:
+    NODE_OPTIONS: '--max-old-space-size=8192'
```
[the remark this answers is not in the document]
habib-deriv: Yes should be fine, locally it takes less than this

.github/workflows/sync-build-cache.yml
habib-deriv: Yes should be fine, locally it takes less than this @adrienne-deriv

.github/workflows/build-to-cloudflare-pages.yml
```diff
@@ -0,0 +1,124 @@
...
+
+                  preview_url=$(wrangler pages deploy $directory --project-name=$projectName --branch=$branch > log.txt 2>&1; echo $?)
+                  echo "------"
+                  cat log.txt
+
+                  branchName=$(echo $branch | sed 's/[\/_]/-/g')
+                  if grep -q "Deployment complete" log.txt; then
+                    echo "preview_url=https://$branchName.deriv-com-playground.pages.dev" > "$GITHUB_OUTPUT"
+                  else
+                    exit 1
```
habib-deriv, line 101: Can we add a message, "deployment failed" ?

adrienne-deriv: Sure I can an additional message, but the `log.txt` is just used to log the progress of running `wrangler pages deploy`, because the Github Actions does not log out the stdout of `wrangler` 😓 , for instance:
### Successful build
*Before adding log file*
Screenshot 2023-08-16 at 1 42 11 PM

*After adding log file*
Screenshot 2023-08-16 at 1 40 25 PM

This is especially useful during debugging deployment errors:
*Before adding log file*
Screenshot 2023-08-16 at 1 43 45 PM

*After adding log file*
Screenshot 2023-08-16 at 1 46 13 PM

habib-deriv: Cool, let's just add the message before exit. LGTM

PR #5878, opened 2023-11-09
Hasan/Translation 00 seo transaltion fix

hasan-deriv:
Changes:

* Typo fix of Options page SEO


review-deriv (commented):
PR description is too short. You can use this format:

```
# Description
This is a summary of the changes. This should describe why do we need this change.
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

PR #5943, opened 2023-11-15
Habib/gtm party town integration

habib-deriv:
Changes:

It works by moving all the script execution to its web worker, thus freeing up the main browser thread to only deal with application code.


package-lock.json
dan-deriv, line 1: @habib-deriv are you sure you want to put `"dev": true` property here?

habib-deriv: Hi @dan-deriv, this is getting auto generated for the modules, those needs this dev flag true

prince-deriv (approved):

PR #6168, opened 2023-12-07
chore: updated cards data

hasan-deriv:
Changes:

-  added id to all cards


markwylde-deriv:
What do these `id`'s map against? Would they be better at actual names describing what the card does? If we add a new card, will that have some arbitary number as the id, or will we shift the `id`'s all the cards after it?

PR #6193, opened 2023-12-12
Prince/DPROD-379/Fetch TrustPilot on build time

prince-deriv:
## Fetch Process

- On pre-build, a script will automatically run to fetch the Trustpilot data.
- The script generates a file at `src/data/trustpilot.json` containing the Trustpilot data. This data is then passed down to the `SocialProof` block from `deriv-com-v2`.


scripts/trustpilot.js
chrisgodfrey-deriv, line 1: I don't think generating a TS file is the way to go here. We should be generating a JSON file, and then mapping data out of that JSON file (as if it were an API call response). Since we're dealing with simple types I don't think the TS adds anything.

[An example of how to do that](https://www.gatsbyjs.com/docs/how-to/sourcing-data/sourcing-from-json-or-yaml/)

scripts/trustpilot.js
```diff
@@ -0,0 +1,6 @@
+const { exec } = require('child_process')
+const { fetchTrustpilotData } = require('./data/trustpilot-generator');
+
+fetchTrustpilotData(() => {
+    exec(`git add src/data/trustpilot.ts`)
```
chrisgodfrey-deriv, line 5: I don't like exec commands like this... it's "magic" stuff that happens (e.g. whats in source control changes on build) and potentially open to misuse.

.husky/pre-commit
```diff
@@ -2,5 +2,7 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run pre-commit-translate
+npm run pre-commit-trustpilot
```
dan-deriv, line 5: I think it should be not a `pre-commit` thing, but a `build` or some type of `pre-build` step. Otherwise, each developer will introduce changes for TrustPilot  but it will not relate to their PRs. 

Also, commit and actual deploy can differ in dates and we will not get the most up to date score.

dan-deriv:
@prince-deriv @chrisgodfrey-deriv I also confirm that we need to change the architecture here. As far as I see it:
1. Pre-build-step: some code fetches scores and generates JSON
3. Build-step: our existing code uses the JSON and bakes in data into corresponding static resources

scripts/trustpilot.js
```diff
@@ -0,0 +1 @@
...312));const _0x205dc0=JSON[_0x2c9a76(0x170)](_0x3c8312,null,0x2);fs['writeFile'](_0x20404c,_0x205dc0,_0x5409d0=>{_0x5409d0&&console['error']('Error\x20writing\x20to\x20the\x20file:',_0x5409d0);});}catch(_0x1d2448){console[_0x2c9a76(0x179)](_0x1d2448);}};function a0_0x3a99(){const _0x353b5c=['path','6UtqFFf','log','5730783cazrKn','deriv.com','1389tgmtgR','&apikey=','\x1b[33m[Trustpilot\x20Data]\x1b[32m:\x20','trustScore','Trustpilot\x20app\x20name\x20or\x20API\x20key\x20is\x20missing','trustpilot.json','resolve','5nMallf','2VnVpQWplNEh','usedForTrustScoreCalculation','68360GrCObY','json','numberOfReviews','11570279CJCnyM','../src/data','120604dNwYri','Network\x20response\x20was\x20not\x20ok:\x20','398724XsrZdY','6243032DfksWs','score','BHaFdLV2g','387MxlJyP','stringify','join','https://api.trustpilot.com/v1/business-units/find?name=','statusText','174BoIAQS','539xbktWx','stars'];a0_0x3a99=function(){return _0x353b5c;};return a0_0x3a99();}function a0_0x403d(_0xe87bc1,_0x4855ab){const _0x3a998d=a0_0x3a99();return a0_0x403d=function(_0x403d36,_0x189c2b){_0x403d36=_0x403d36-0x15d;let _0x461a02=_0x3a998d[_0x403d36];return _0x461a02;},a0_0x403d(_0xe87bc1,_0x4855ab);}fetchTrustpilotData();
```
dan-deriv, line 1: This can be refactored, in a way that we get api-key from environment variables. For that github-actions will need to be changed. But Im not aware of how it's completely organized in this project. 

Consequently, you will not need to encrypt the script at all. 

cc: @markwylde-deriv @habib-deriv @mohsen-deriv @chrisgodfrey-deriv

.gitignore
```diff
@@ -61,6 +61,9 @@ typings/
 .env.development
 .env.production
 
+# build time data
+scripts/trustpilot-src.js
```
dan-deriv, line 65: I do not see where this file is used. Are you sure we need to gitignore?

scripts/trustpilot.js
chrisgodfrey-deriv: I agree,... I prefer not to encode like this and tokenise the API key. The API key should be leveraged an environment variable.

The problem with encoding the code this way is that maintenance is difficult which makes the DevEx poor. 

We should have a "Staging" key and a "Production" key also. For local environment we can just mock the data.

markwylde-deriv: What does this "production" key give users access. There's no such thing as a secret on the frontend.

Don't trustpilot give you a tag/code snippet to include on the site so we don't have to do things like this?

.github/workflows/generate-preview-link.yml
```diff
@@ -129,6 +129,7 @@ jobs:
                   GATSBY_RUDDERSTACK_PRODUCTION_KEY: ${{ secrets.GATSBY_RUDDERSTACK_PRODUCTION_KEY }}
                   GATSBY_STRAPI_TOKEN: ${{ secrets.GATSBY_STRAPI_TOKEN }}
                   GATSBY_GOOGLE_TAG_MANAGER_TRACKING_ID: ${{ secrets.GATSBY_GOOGLE_TAG_MANAGER_TRACKING_ID }}
+                  TRUSTPILOT_API_KEY: ${{ secrets.TRUSTPILOT_API_KEY }}
```
habib-deriv, line 132:
```suggestion
                  The enviroment needs to have GATSBY_ prefix both in repo secret and YML files
```

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+require('dotenv').config();
+
+const fetchTrustpilotData = async () => {
+    const fileDir = path.resolve(__dirname, '../src/data')
+    const fileName = 'trustpilot.json'
+    const filePath = path.join(fileDir, fileName)
+ 
+    try {
+        const appName = 'deriv.com'
+        const apiKey = process.env.TRUSTPILOT_API_KEY
```
habib-deriv, line 12:
```suggestion
        const apiKey = process.env.TRUSTPILOT_API_KEY
```

Same here

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+            numberOfReviews:
+                result.numberOfReviews?.usedForTrustScoreCalculation.toLocaleString() || '',
+        }
+
+        const fileContent = JSON.stringify(tpData,null,2)
+
+        fs.writeFile(filePath, fileContent, (error) => {
+            if (error) {
+              // Handle errors
+              console.error('Error writing to the file:', error);
```
habib-deriv, line 38: Can we show error message or TP logo here for this, instead of console?

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+ 
+    try {
+        const appName = 'deriv.com'
+        const apiKey = process.env.TRUSTPILOT_API_KEY
+     
+        if (!appName || !apiKey) {
+            throw new Error('Trustpilot app name or API key is missing')
+        }
+
+        const url = `https://api.trustpilot.com/v1/business-units/find?name=${appName}&apikey=${apiKey}`
```
habib-deriv, line 18: Can we move this url to constant.ts file?

gatsby-node.js
```diff
@@ -3,11 +3,28 @@ const language_config = require(`./i18n-config.js`)
...
 
+const fetchTrustpilotData = () => {
+    // Trustpilot on-build data fetching
+    const startTime = Date.now()
+
+    try {
+        execSync('node scripts/trustpilot.js')
+        const endTime = Date.now()
+        const timeSpentInSeconds = (endTime - startTime) / 1000
+        console.log(
```
habib-deriv, line 17: We can remove this

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+        const response = await fetch(url)
+
+        if (!response.ok) {
+            throw new Error(`Network response was not ok: ${response.statusText}`)
+        }
+
+        const result = await response.json()
+        const tpData = {
+            stars: result.score?.stars || 0,
```
habib-deriv, line 27: Instead of 0, can we have null, in case the api fails we can show a TP image maybe?

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+            trustScore: result.score?.trustScore || 0,
```
habib-deriv, line 28: same

scripts/trustpilot.js
```diff
@@ -0,0 +1,48 @@
...
+
+        fs.writeFile(filePath, fileContent, (error) => {
+            if (error) {
+              // Handle errors
+              console.error('Error writing to the file:', error);
+            }
+          });
+       
+    } catch (error) {
+        console.log(error)
```
habib-deriv, line 43: Can we handle this error with error message or something?

scripts/trustpilot.js
habib-deriv: Can ignore

scripts/trustpilot.js
prince-deriv: No @habib-deriv this works as a data. We have a default file the contains static data. If fetching fails that data file won't change and thus will always show trustpilot regardless of the result of this fetch call.

habib-deriv (approved):
LGTM

mohsen-deriv (approved):

PR #6246, opened 2023-12-15
Prince/update trustpilot build order

prince-deriv:
## Changes

- updated order of fetching trustpilot data. 
- the data should be fetched before building the pages. 
- fetch script moved to `onPreInit`


habib-deriv (approved):
LGTM

PR #6462, opened 2024-01-09
Pr 2169 quill layout fixes

hasan-deriv:
Changes:

- Bring back ubuntu fonts. Since, quill tokens only have bold variant of ubuntu font
- Fixed the background color of floating button
- Updated tailwind config to let tailwind scan block, components and quill-design classes. In this way, tailwind won't create any duplicate class and this fixed our css piority issue.


mitra-deriv (approved):
LGTM 🌟

PR #6657, opened 2024-01-16
Prince/Update fetching total tp reviews

prince-deriv:
## Changes

- updated data being use to show total trustpilot reviews/


review-deriv (commented):
PR description is too short. You can use this format:

```
# Description
This is a summary of the changes. This should describe why do we need this change.
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

hasan-deriv (approved):

PR #6904, opened 2024-01-29
Prince/ added translation-fix command for conflict fixing

prince-deriv:
**Updates:**

- added a new script `npm run translation:fix` - this is to resolve conflicts between `master` and `translation` branches

### DIFF
--- a/scripts/handle-translations.js
+++ b/scripts/handle-translations.js
@@ -122,6 +122,7 @@ const runProcess = (data) => {
                 `git merge -X theirs --commit master -m "chore: $(git status --short)"`,
                 callback,
             )
+            break
         default:
             break
     }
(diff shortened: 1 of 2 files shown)


scripts/handle-translations.js
```diff
@@ -118,6 +117,11 @@ const runProcess = (data) => {
                 callback,
             )
             break
+        case 'merge-master-translation':
+            exec(
+                `git merge -X theirs --commit master -m "chore: $(git status --short)"`,
+                callback,
+            )
```
mitra-deriv, line 124: do we need a break here @prince-deriv ? or we wanna fall through the default? 🙏

mitra-deriv (approved):

scripts/handle-translations.js
prince-deriv: yes we need a break.

mitra-deriv (approved):

PR #6965, opened 2024-02-01
Hasan/DPROD-3065/lazy loading

hasan-deriv:
Changes:

- added lazy loading for live market, trade types and payment section

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release

### DIFF
--- a/src/features/pages/home/index.tsx
+++ b/src/features/pages/home/index.tsx
@@ -1,20 +1,20 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
-import React from 'react'
+import React, { lazy, Suspense } from 'react'
 import { PageLayout } from '@deriv-com/components'
 import QuillLayout from 'features/components/quill/quill-layout'
-import LiveMarketSection from './live-pricing-migration'
 import TwentyYearsStrong from './years'
 import StatSection from './stats'
 import UserFriendlyPlatforms from './user-platfroms'
-import TradeTypeSection from './trade-type'
 import HomeHero from './hero'
-import FastPaymentSection from './fast-payment'
 import CTA from './cta'
 import StartTradingSteps from './start-trading-steps'
 import TrustpilotSection from './trustpilot'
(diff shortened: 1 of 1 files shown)


src/features/pages/home/index.tsx
```diff
@@ -1,20 +1,20 @@
...
-import TradeTypeSection from './trade-type'
 import HomeHero from './hero'
-import FastPaymentSection from './fast-payment'
 import CTA from './cta'
 import StartTradingSteps from './start-trading-steps'
 import TrustpilotSection from './trustpilot'
 import { useOpenLiveChat } from 'components/hooks/use-open-live-chat-redirection'
 import MainRowNavigation from 'features/components/templates/navigation/main-nav'
 import MainFooter from 'features/components/templates/footer'
+const FastPaymentSection = lazy(() => import('./fast-payment'))
```
wojciech-deriv, line 15: how do we know that this library was actually excluded from main bundle? 
(reason I'm asking is that I used to implement similiar patterns in different project and commonly I have been stumbling upon the fact that some random tiny component was still importing something from file intended to be lazy loaded and thus preventing webpack from removing it from main bundle)

dan-deriv: it was not, because we compile ALL code into a single file. we do not have chunks as they create too much overhead for the network and almost no win with size.

wojciech-deriv: so if its not excluded from main bundle, then whats the point of lazy loading?

wojciech-deriv: IMO lazy loading does make sense only if given resource is not already included in my bundle

dan-deriv:
Closing in favor of https://github.com/binary-com/deriv-com/pull/6966/

PR #7054, opened 2024-02-12
mayuran/DPROD-2759/ws performance

mayuran-deriv:
Changes:

-   ...

## Type of change

-   [X] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


gatsby-browser.js
```diff
@@ -44,9 +44,13 @@ const hacker_one_url = 'https://hackerone.com/deriv?type=team'
 if (is_browser) {
     if (window.location.pathname.includes(bug_bounty_path)) window.location.href = hacker_one_url
 }
+let GlobalProviderInstance = null;
 
 export const wrapRootElement = ({ element }) => {
-    return <GlobalProvider>{element}</GlobalProvider>
+    if (!GlobalProviderInstance) {
```
wojciech-deriv, line 50: Gatsby does do not suggest its needed - any particular reason for that "if" statament? 

Also, subsequent calls will return "element" from the first call - which I do not belive is correct, so that needs to be somehow improved (ideally: investigate the problem deeper / follow recommendation from gatsby when implementing stuff). Happy to jump on call / pair on that. 

https://www.gatsbyjs.com/docs/reference/config-files/gatsby-ssr/#wrapRootElement

dan-deriv:
@mayuran-deriv pls rename title

gatsby-browser.js
dan-deriv: I do not follow what is the effect of this thing, as well as how it relates to the web-sockets call  and a cookie.

PR #7062, opened 2024-02-13
Eduard/DPROD-3367/Override social-buttons urls according to regional settings

eduard-deriv:
Changes:

-   Override with lang and GrowthBook integration

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/features/components/templates/footer/data.tsx
```diff
@@ -498,91 +499,119 @@ export const socialButtonsROW = [
...
+        'data-cy': SocialProviders.Instagram,
+        'aria-label': SocialProviders.Instagram,
     },
     {
         icon: <StandaloneXTwitterIcon iconSize="md" fill="#000000b8" />,
         href: 'https://twitter.com/derivdotcom/',
-        'data-cy': 'twitter',
-        'aria-label': 'twitter',
+        'data-cy': SocialProviders.Twitter,
+        'aria-label': SocialProviders.Twitter,
```
mitra-deriv, line 515: Why do we need to change these? if it's a need Please help to fix the test failure also in cypress due to changes in data-cy

src/components/hooks/use-app-config.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Analytics } from '@deriv-com/analytics'
+
+export enum SocialProviders {
+    Facebook = 'facebook',
+    Twitter = 'twitter',
+    Instagram = 'instagram',
+    Telegram = 'telegram',
+    Youtube = 'youtube',
+    Linkedin = 'linkedin',
+}
```
hasan-deriv, line 11:
```suggestion
export const SocialProviders =  {
    Facebook: 'facebook',
    Twitter: 'twitter',
    Instagram: 'instagram',
    Telegram: 'telegram',
    Youtube: 'youtube',
    Linkedin: 'linkedin',
} as const
```

hasan-deriv: it's better to not to use enum

hasan-deriv (changes requested):

src/features/components/templates/footer/index.tsx
```diff
@@ -46,4 +49,22 @@ export const MainFooter = () => {
...
+                : button,
+        )
+
+    let buttons = is_career ? socialButtonsCareers : is_eu ? socialButtonsEU : socialButtonsROW
+
+    buttons = buttons.filter((button) => !!config[button?.['aria-label']])
+    buttons = overrideWithLang(buttons)
+
+    return buttons
+}
```
hasan-deriv, line 68: Can you explain this? What is happening here?

mitra-deriv: same I'm so confused, I think we can make this simpler what is causing it to add this much complexity?

eduard-deriv: we need to check region (eu or row), special langs has special links (need to change only 3 social links, not 6) for both row and eu regions, need to filter by growthBook config (is it enabled or not), also not forget to check and change links for career page. thats why you think that is so complicated.

habib-deriv:
Hi @eduard-deriv , can we remove growthbook integration from here?

src/features/components/templates/footer/index.tsx
dan-deriv: @hasan-deriv @mitra-deriv good question. This could have been explained in the PR description. Effectively, there is no complexity here, but let me explain how the algorithm works:

1) If a user is on a career-page -- he gets some links, if on other pages -- standard links. 

2) The links that are going to be returned first get checked if they are enabled in the config (upcoming feature)

3) Then links are overridden with some special-regional links (current feature request).

dan-deriv:
> Hi @eduard-deriv , can we remove growthbook integration from here?

@habib-deriv Is it really needed? Deadline for dynamic configurations is this week, but Firebase is not yet released. So, this config shall be removed as soon as Firebase comes in.

habib-deriv:
> > Hi @eduard-deriv , can we remove growthbook integration from here?
> 
> @habib-deriv Is it really needed? Deadline for dynamic configurations is this week, but Firebase is not yet released. So, this config shall be removed as soon as Firebase comes in.

One visible problem here, we have multiple facebook accounts. Global one, Careers, different fb account for different language. So we can't have feature flag like this. Let's say we need to hide deriv Arabic fb account but need to keep other accounts, then this won't work . So I think we need to merge @fasihali-deriv's PR that will handle all these scenerios

src/components/hooks/use-app-config.tsx
habib-deriv, line 1: I think we can delete this file?

habib-deriv (approved):

habib-deriv (dismissed):

PR #7070, opened 2024-02-15
mayuran/DPROD-2759/ws_improvements

mayuran-deriv:
stop ping call
stop calling WS calls twice
-   ...

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Update feature
-   [ ] Refactor code
-   [ ] Translation to code
-   [ ] Script configuration
-   [ ] Improve performance
-   [ ] Style only
-   [ ] Dependency update
-   [ ] Documentation update
-   [ ] Release


src/common/websocket/index.ts
```diff
@@ -34,22 +33,59 @@ export class ApiManager {
...
                 const app_id = getAppId()
                 const websocket_connection_url = `${socket_url}?app_id=${app_id}&l=${language}&brand=${brand_name.toLowerCase()}`
 
                 this.socket = new WebSocket(websocket_connection_url)
+                this.setReadyStateInSessionStorage(this.socket?.readyState)
             }
             this.derivApi = new DerivAPIBasic({ connection: this.socket })
-            this.registerKeepAlive()
+            this.socket.addEventListener('open', () => {
+                console.log('WS connected.')
```
mayuran-deriv, line 58: kept console logs for QA tracing purpose, later will remove console logs.

habib-deriv (approved):
LGTM

prince-deriv (approved):

src/common/websocket/index.ts
```diff
@@ -34,22 +33,59 @@ export class ApiManager {
         return ApiManager.instance
     }
 
+    public static readonly READY_STATE_KEY = 'websocket_ready_state'
+
+    private setReadyStateInSessionStorage(state: number) {
```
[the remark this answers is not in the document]
mayuran-deriv: hi @wojciech-deriv ,
actually i needed a global state in this call to check the ready_state of WS calls , and in deriv APP they are using MOBX to achieve this   since we don't have  global state like redux or mobx i used session storage.

src/common/websocket/index.ts
```diff
@@ -34,22 +33,59 @@ export class ApiManager {
...
+            this.socket.addEventListener('open', () => {
+                console.log('WS connected.')
+                this.setReadyStateInSessionStorage(this?.socket?.readyState)
+            })
+
+            this.socket.addEventListener('close', () => {
+                console.log('WS closed')
+                this.derivApi.disconnect()
+                this.ready = null
+                sessionStorage.removeItem(ApiManager.READY_STATE_KEY)
```
dan-deriv, line 66: You should have kept the implementation of your storage inside the `setReadyStateInSessionStorage`. With your approach, if you change it from session-storage to local-storage you need to update all places. 

So, if you had something like 
```
// PSEUDO

private setReadyStateInSessionStorage(key: string, value: number) {
   if (value == null) sessionStorage.removeItem(key)
   else sessionStorage. setItem(key, value)
}
```

you should both _set_ and _unset_ things via this function.

src/components/hooks/useWS.ts
```diff
@@ -15,13 +16,29 @@ const useWS = <T extends TSocketEndpointNames>(name: T) => {
...
         async (data?: Parameters<typeof apiManager.augmentedSend<T>>[1]) => {
             setIsLoading(true)
-            try {
-                const response = await apiManager.augmentedSend(name, data)
-                setData(response[name] as TSocketResponseData<T>)
-            } catch (e) {
-                setError(e)
-            } finally {
-                setIsLoading(false)
+            const readyState = parseInt(sessionStorage.getItem('websocket_ready_state'))
```
dan-deriv, line 19: Constant is not used

src/components/hooks/useWS.ts
```diff
@@ -15,13 +16,29 @@ const useWS = <T extends TSocketEndpointNames>(name: T) => {
...
+                    const response = await apiManager.augmentedSend(name, data)
+                    setData(response[name] as TSocketResponseData<T>)
+                } catch (e) {
+                    setError(e)
+                } finally {
+                    setIsLoading(false)
+                }
+            } else {
+                if (isBrowser()) {
+                    const currentLanguage = getLanguage() ?? 'en'
```
dan-deriv, line 31: Can't `getLanguage()` return a default lang? Why?
I think that instead of such checks you should be able to return a default.

mayuran-deriv: its old method which we are using to get language  so i didn't change the method

src/components/hooks/useWS.ts
```diff
@@ -15,13 +16,29 @@ const useWS = <T extends TSocketEndpointNames>(name: T) => {
...
             setIsLoading(true)
-            try {
-                const response = await apiManager.augmentedSend(name, data)
-                setData(response[name] as TSocketResponseData<T>)
-            } catch (e) {
-                setError(e)
-            } finally {
-                setIsLoading(false)
+            const readyState = parseInt(sessionStorage.getItem('websocket_ready_state'))
+            if (readyState === 1 || readyState === 0) {
```
dan-deriv, line 20: What else it can be?

Why you do not consider both undef, and 0 as a candidate for reconnect?

mayuran-deriv: while connecting the state will be set to 0 
after connected it will be 1 
so no need to check for undef

src/store/website-status-context.tsx
```diff
@@ -30,8 +33,20 @@ export const WebsiteStatusProvider = ({ children }: WebsiteStatusProviderProps)
...
+            statusDataPromise = new Promise((resolve) => {
+                statusDataPromiseResolve = resolve
+            })
+            send()
+        } else if (statusDataPromise) {
+            statusDataPromise.then((data) => {
+                setData(data)
+            })
+        }
+    }, [send, wsData])
```
dan-deriv, line 49: There is some hack here. I do not understand all this ceremony. Moreover,
1. Your effect will not be regenerated if statusDataPromise or statusDataPromiseResolve changed
2. It is somehow dependent on the real-time code-flow, eg there is a chance that you call statusDataPromiseResolve(wsData) for a _null_

Seems like this block needs to be rewritten.

src/components/hooks/useWS.ts
dan-deriv: You can improve it. Why not.

src/common/websocket/index.ts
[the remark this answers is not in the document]
mayuran-deriv: hi @wojciech-deriv changed  too static variable

src/common/websocket/index.ts
mayuran-deriv: removed session storage so resolving this.

src/components/hooks/useWS.ts
mayuran-deriv: removed session storage so resolving this

src/store/website-status-context.tsx
mayuran-deriv: here we are using promise and resolve to if two request going on a same time  we are detecting the in progress request and avoiding the new request. 
and for null we are checking if(wsData) so null values wont hit  statusDataPromiseResolve(wsData) this line

src/components/hooks/useWS.ts
```diff
@@ -15,13 +16,29 @@ const useWS = <T extends TSocketEndpointNames>(name: T) => {
...
+                } catch (e) {
+                    setError(e)
+                } finally {
+                    setIsLoading(false)
+                }
+            } else {
+                if (isBrowser()) {
+                    const currentLanguage = getLanguage() ?? 'en'
+                    apiManager
+                        .reconnectIfNotConnected(currentLanguage)
```
wojciech-deriv, line 33: question: if I understand correctly, if its already connected, it will send using ".augumentedSend", if its not connected already, it will connect and then send using just ".send()". Does that matter / is it accidentall difference / purposefull difference?

also, would it be easier to do slightly differenct logic:

a) check if its connected at the top of the function, if its not, reconnect,
b) just do exactly the same which is doing now, without extra conditions?

wojciech-deriv (approved):

src/components/hooks/useWS.ts
wojciech-deriv: thats where the high test coverage helps :) to increase confidence and thus enable refactors like that ;)
so developers are not worried of changing stuff as they know its gonna be caught by tests

src/components/hooks/useWS.ts
mayuran-deriv: so something like this 
`    const send = useCallback(
        async (data?: Parameters<typeof apiManager.augmentedSend<T>>[1]) => {
            setIsLoading(true)

            // Check if WebSocket connection is available
            if (parseInt(ApiManager.readyState) !== 1 && parseInt(ApiManager.readyState) !== 0) {
                if (isBrowser()) {
                    const currentLanguage = getLanguage() ?? 'en'
                    try {
                        // Connect to WebSocket if not connected
                        await apiManager.reconnectIfNotConnected(currentLanguage)
                    } catch (e) {
                        setError(e)
                        setIsLoading(false)
                        return
                    }
                } else {
                    setIsLoading(false)
                    return
                }
            }

            // Proceed with sending data
            try {
                const response = await apiManager.augmentedSend(name, data)
                setData(response[name] as TSocketResponseData<T>)
            } catch (e) {
                setError(e)
            } finally {
                setIsLoading(false)
            }
        },
        [name],
    )`

mayuran-deriv: something like this 


    const send = useCallback(
        async (data?: Parameters<typeof apiManager.augmentedSend<T>>[1]) => {
            setIsLoading(true)

            // Check if WebSocket connection is available
            if (parseInt(ApiManager.readyState) !== 1 && parseInt(ApiManager.readyState) !== 0) {
                if (isBrowser()) {
                    const currentLanguage = getLanguage() ?? 'en'
                    try {
                        // Connect to WebSocket if not connected
                        await apiManager.reconnectIfNotConnected(currentLanguage)
                    } catch (e) {
                        setError(e)
                        setIsLoading(false)
                        return
                    }
                } else {
                    setIsLoading(false)
                    return
                }
            }

            // Proceed with sending data
            try {
                const response = await apiManager.augmentedSend(name, data)
                setData(response[name] as TSocketResponseData<T>)
            } catch (e) {
                setError(e)
            } finally {
                setIsLoading(false)
            }
        },
        [name],
    )

meenakshi-deriv (approved):

